PR #1, opened 2026-03-27
test: Add basic playwright tests opening examples

juliusknorr:
- **test: Add basic playwright tests opening examples**

```
E2E_BASE_URL=http://localhost:8080 npm run test:headed

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -442,3 +442,67 @@ jobs:
       #    name: packages-${{ github.sha }}
       #    path: dist/packages/
       #    retention-days: 30
+
+  e2e:
+    runs-on: self-hosted
+    needs: combine
+    # Only run when the image was actually pushed (main branch or version tags)
+    if: github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/')
+    permissions:
+      contents: read
+      packages: read
+
+    steps:
+      - name: Restore cache
+        uses: actions/cache/restore@v4
+        with:
+          path: "${{ github.workspace }}"
+          key: ${{ runner.os }}-build-${{ github.sha }}
+
+      - name: Log in to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Determine image tag
+        id: image
+        run: |
+          if [[ "${{ github.ref }}" == refs/tags/* ]]; then
+            echo "ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:latest" >> $GITHUB_OUTPUT
+          else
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: .github/workflows/build.yml [hunk 1]
```diff
@@ -377,3 +396,72 @@ jobs:
               create_manifest "$repo" "${REF}" "latest"
             done
           fi
+
+  e2e:
+    runs-on: self-hosted
```
rikled, line 435: Do we still want to run this on self-hosted?

rikled (approved):
looks good to me. 

Some claude comments:

**Overall this is solid, well-engineered work**

### Issues to address

**[High] Fork PRs will fail and can't be fixed without `pull_request_target`.**
A `pull_request` from a fork gets a read-only `GITHUB_TOKEN` and no secrets, so the `clone` job's `EURO_OFFICE_MIRROR_TOKEN` is empty and the GHCR push (`packages: write`) can't run. Fine if PRs are always same-repo branches, but external contributors will hard-fail. Worth a deliberate decision/comment.

**[High] No `concurrency` guard + fixed container name and host port.**
`global-setup.ts` hardcodes `CONTAINER_NAME = 'euro-office-e2e'` and `HOST_PORT = 8988`. If the self-hosted runner ever runs two e2e jobs at once, they collide on the port and the idempotent `docker rm -f` can kill the other run's container. Suggest:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

**[Medium] `GITHUB_TOKEN` may not be able to delete org-level package versions.**
In `e2e-image-cleanup.yml`, deleting an org container package version via the default token usually requires the repo to be granted admin on the package (or a PAT). If not, cleanup silently no-ops and `pr-<n>` tags accumulate. Worth verifying on a real close event.

**[Medium] Fragile coupling between `HOST_PORT` and the config default.**
Playwright evaluates `playwright.config.ts` before `globalSetup` runs, so the `process.env.E2E_BASE_URL` set in setup never reaches `use.baseURL`. It only works because the config fallback (`http://localhost:8988`) equals `HOST_PORT`. Change one without the other and `baseURL` breaks silently — consider deriving both from one source.

**[Low] Possible startup race on the example app.**
`/healthcheck` reflects docservice readiness, not the separate `ds-example` process, so the first `page.goto('/example/')` could occasionally hit a 502 before it's up. The 15s `toHaveTitle` retry + `retries: 1` mostly absorb this, but it's a likely flake source.

### Minor nits
- The tests assume clicking `a.try-editor.word` opens the editor in a **new tab** (`waitForEvent('page')`); if it ever opens same-tab, the test hangs to timeout.
- `test:url` in `e2e/package.json` is identical to `test`.
- The e2e `if:` hardcodes `refs/heads/main` while `manifest` uses `repository.default_branch` — harmless unless the default branch is renamed.

None of these block an internal-only, single-runner setup, but I'd add the concurrency guard and confirm the cleanup token before relying on it.

PR #21, opened 2026-03-30
feat: :hammer: Update license processing script

moodyjmz:
Better script for license processing. Retains the 'dry run feature' and takes into account more edge cases (formatting, addresses). Uses simpler string matching - also covers UNIX and Windows line endings.


juliusknorr:
Let's close this for now as we'll see if the new upstream license can be kept.

PR #25, opened 2026-03-30
fix: broken double quote in pgrep call. 

LMinauda:
Nested double double quotes around systemd and supervisord in pgrep calls.


rikled:
Interesting, this bug was already present in the OnlyOffice code. Thanks :)

PR #40, opened 2026-03-31
feat: readd arm build workflow, broke after merge of core

rikled:
After the massive changes of core for the desktop editors, arm builds broke. This PR readds arm builds to the github workflow

This requires
- [x] v8
- [x] boost

to be precompiled and uploaded to the binary upstream in core.

depends on: https://github.com/Euro-Office/core/pull/57


rikled:
I downgraded emscripten to 3.1.48 for which there does not seem to exist an arm image. The downgrade is probably not necessary and we can maybe again go to version 5.0.0 but not sure.

PR #43, opened 2026-04-01
chore(workflow): update updatemirror to add contributors list to REAdme

alphaleadership:
…ME.md


juliusknorr:
Thanks for the contribution! We're going to close this one though.

We want to keep the README focused as a quick entrypoint — installation, usage, the essentials. A full list of contributors will be too much there and works against that goal. Also the script does not paginate and breaks the existing mirror update script.

Feel free to open a discussion if you'd like to revisit this!

PR #71, opened 2026-04-14
fix(build): bump CACHE_BUST to 3 and update core submodule

chrip:
- CACHE_BUST 2→3 forces a new wasm-build-cache-3 Docker mount, which causes core-wasm stage to fully recompile fonts.wasm and fonts.js from the updated module_js.js (onRuntimeInitialized fix).
- Update core submodule pointer to include the FT_* timing fix.


rikled:
done already :+1:

PR #75, opened 2026-04-17
Feat/support android dev ips

### LINKED ISSUE #74: Support Android local dev IP + more
Android uses 10.0.2.2 locally. This is not in the trusted domains - it needs adding to /develop/docker-compose.yml

```sh
  NEXTCLOUD_TRUSTED_DOMAINS: 172.18.0.1 10.0.2.2                                                                              
```

10.0.2.2 is analogous to localhost.

moodyjmz:
Closes https://github.com/Euro-Office/DocumentServer/issues/74

Testing on desktop browsers is fine. Testing on Mac + simulated iPhone is also fine. As is iPhone in Nextcloud App.

However, Android is not fine. Android maps 10.0.2.2 to localhost, this doesn't work.

We have to do two things

1) Ensure 10.0.2.2 is allowed

2) Set the DocumentServerUrl to something accessible on the local net as it gets set to the OO iframe

Impl:

1) add Android dev IP to trusted and ensure that config doesn't overwrite this. (also removes a previous value there that was being overwritten)

~~2) As we need the actual IP, get this and also introduce a way to keep this updated - when networks change without having to pull docker all down etc - as part of make local this happens, and also adds make refresh to allow quick updates if needed.~~ Made nicer approach using `make mobile` - one time docker update, then it is there

This way, everything works :)

### Testing

Inside DocumentServer/develop:

`docker down -v` - This is remove existing EO so, you might want to do this on a version you aren't dev'ing on. **One time thing.**

`make local` - runs as normal

`make mobile` - uses localhost for desktop, but will also work for Android devices simulated via Android Studio (can use the IP make mobile provides, or 10.0.2.2 - the one Android Studio maps to localhost)

Switch between the two to your heart's content. Testing as appropriate.

### DIFF
--- a/develop/Makefile
+++ b/develop/Makefile
@@ -1,7 +1,21 @@
 PRODUCT_VERSION ?= $(shell cat ../VERSION)
 export PRODUCT_VERSION
 
+# Host LAN IP so phones/emulators/LAN devices can reach DocumentServer and
+# Nextcloud via the desktop's routable address rather than localhost.
+ifeq ($(shell uname -s),Darwin)
+	DETECTED_HOST_LAN_IP := $(shell ipconfig getifaddr en0 2>/dev/null || ipconfig getifaddr en1 2>/dev/null)
+else
+	# 1.1.1.1 is a dummy public target — `ip route get` just asks the kernel
+	# which local source IP it would use, without actually sending packets.
+	DETECTED_HOST_LAN_IP := $(shell ip route get 1.1.1.1 2>/dev/null | awk '{print $$7; exit}')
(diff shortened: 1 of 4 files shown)


moodyjmz:
switched back to draft to test more - cache... need to stress test it

moodyjmz:
ah, it is the IP rotation, so behaving as expected, however I have a little think

moodyjmz:
Switching to draft as I don't like the make local changing. These changes should be independent of that.

AndyScherzinger (approved):

PR #91, opened 2026-05-11
Create default PostgreSQL cluster (required before starting the service)

efemiydi:
Fix PostgreSQL initialization in Docker build: add missing cluster creation (pg_createcluster), otherwise service start fails with "No PostgreSQL clusters exist"


rikled:
Thanks for the PR :). We're currently working on several changes to the standalone Docker image. I'll merge this once that's done.

rikled:
Thanks again for the PR. I added your changes here: https://github.com/Euro-Office/DocumentServer/commit/79e5e73488ae64868785fbd2bce2032006a45c95

Since there were significant changes to the project recently I could not easily merge it. So I'll close here.

PR #93, opened 2026-05-12
Feat/enhance make dx

moodyjmz:
Make dev setup more robust and provide more useful feedback

Doesn't delete data volumes when switching to support different NC versions

Provides feedback if install is taking a long time

Provides feature to wipe targets

Adds `make next-mobile` - akin to `make next`

### DIFF
--- a/develop/Makefile
+++ b/develop/Makefile
@@ -35,18 +35,19 @@ ifeq ($(HOST_LAN_IP),)
 	@echo "   For Android / physical device testing: make mobile"
 else
 	@echo "   Mode: mobile / LAN"
-	@echo "   Host LAN IP:   $(HOST_LAN_IP)"
+	@echo "   Nextcloud:     http://$(HOST_LAN_IP):8081/"
 	@echo "   On IP change:  make refresh-urls"
 endif
+	@echo "   Volume:    nc_data_$${NC_VERSION:-latest}  (wipe with: make wipe-nc NC_VERSION=$${NC_VERSION:-latest})"
 	@echo ""
 	docker compose up -d
 	@if [ "$$(docker inspect --format '{{.State.Status}}' nextcloud 2>/dev/null)" != "running" ]; then \
 	  if docker compose logs nextcloud 2>&1 | grep -q "version of the data"; then \
 	    echo ""; \
-	    echo "   WARNING: Nextcloud failed to start — your local nextcloud:latest image is"; \
+	    echo "   WARNING: Nextcloud failed to start — your local nextcloud:$${NC_VERSION:-latest} image is"; \
 	    echo "   behind the data volume. Pull the latest image and retry:"; \
 	    echo ""; \
-	    echo "      docker pull nextcloud:latest && make"; \
+	    echo "      docker pull nextcloud:$${NC_VERSION:-latest} && make"; \
 	    echo ""; \
 	    exit 1; \
 	  fi; \
--- a/develop/Makefile
+++ b/develop/Makefile
@@ -40,6 +40,17 @@ else
 endif
 	@echo ""
 	docker compose up -d
+	@if [ "$$(docker inspect --format '{{.State.Status}}' nextcloud 2>/dev/null)" != "running" ]; then \
+	  if docker compose logs nextcloud 2>&1 | grep -q "version of the data"; then \
+	    echo ""; \
+	    echo "   WARNING: Nextcloud failed to start — your local nextcloud:latest image is"; \
+	    echo "   behind the data volume. Pull the latest image and retry:"; \
+	    echo ""; \
+	    echo "      docker pull nextcloud:latest && make"; \
+	    echo ""; \
+	    exit 1; \
+	  fi; \
+	fi
 	@$(MAKE) --no-print-directory refresh-urls
 	@$(MAKE) --no-print-directory print-nc-version
 	@docker compose exec eo bash || true
--- a/develop/Makefile
+++ b/develop/Makefile
@@ -97,13 +97,27 @@ build:
 refresh-urls:
 	@start=$$(date +%s); \
 	  while ! docker compose exec -T eo curl -sf http://localhost/healthcheck >/dev/null 2>&1; do \
+	    if [ "$$(docker inspect --format '{{.State.Status}}' eo 2>/dev/null)" != "running" ]; then \
+	      printf "\r[eo]        failed — check: docker compose logs eo\n"; \
+	      exit 1; \
+	    fi; \
 	    printf "\r[eo]        waiting     (%ds)   " $$(($$(date +%s) - start)); \
 	    sleep 2; \
 	  done; \
 	  printf "\r[eo]        ready        (%ds)        \n" $$(($$(date +%s) - start))
 	@start=$$(date +%s); \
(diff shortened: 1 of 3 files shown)


rikled:
Do we even need “make local” anymore? I find it a bit confusing that we have NC_VERSION and NC_BRANCH. I think julius's image should be enough, right? Maybe I'm misunderstanding something. What do you think?

Other than that, it looks good.

chrip (changes requested):
## Summary

This PR enhances the `develop/` Makefile and docker-compose setup with:

1. **`make mobile`** — auto-detects host LAN IP, injects it into Nextcloud so Android emulators and physical LAN devices can reach the DocumentServer editor.
2. **`make next` / `make next-mobile`** — tests against unreleased Nextcloud versions via `nextcloud-docker-dev` image, with per-branch data volumes so state is preserved across switches.
3. **`make refresh-urls`** — idempotent post-boot hook that rewrites `DocumentServerUrl` and `trusted_domains` without recreating the container.
4. **`make wipe-nc` / `make wipe-next`** — clean data volumes per version/branch.
5. **`NC_VERSION` pinning** — `NC_VERSION=33 make local` pins to a specific NC major version with its own data volume.

## What Worked

| Target | NC Version | Mode | Result |
|--------|-----------|------|--------|
| `make local` | 33.0.3 | localhost | Passed |
| `make mobile` | 33.0.3 | LAN IP | Passed (tested on iPhone) |
| `make next` | 35.0.0 dev | localhost | Passed |
| `make next-mobile` | 35.0.0 dev | LAN IP | Passed (tested on iPhone) |
| DOCX editing | — | — | Passed |
| `wipe-next` | — | — | DOCX deleted correctly |
| `wipe-nc` | — | — | DOCX deleted correctly |

All four targets are functional. The LAN IP injection works correctly on iOS. Both wipe targets correctly destroy data volumes.

## Issues Found

### 1. `refresh-urls` timeout on `make next` first boot (BUG)

**File:** `Makefile:130-145`

`refresh-urls` waits for `eurooffice` app to be installable by polling `occ config:app:get eurooffice DocumentServerUrl`. But `nextcloud-docker-dev` takes several minutes on first boot to clone the NC source and install into the volume. The polling loop has **no exit timeout** — it runs forever.

The "still installing" banner after 1000s confirms progress but doesn't help if you walk away.

**Suggestion:** Add a hard timeout (e.g., 600s) with a clear message: `"Timed out after 10m — the first boot of nextcloud-docker-dev is slow. Wait for it to finish, then run make refresh-urls."`

### 2. `wipe-nc` / `wipe-next` silently fail when container is running (BUG)

**File:** `Makefile:98` and `Makefile:106`

Both targets use `docker compose stop` which does not release the volume mount. When the container is running, the volume is still attached, so `docker volume rm` fails silently:

```
Removing volume nc_data_master...
Volume not found — nothing to remove.
```

The volume still exists — it was never actually removed. Tested by creating a DOCX file, running wipe, then verifying the file was still present in the volume.

Verified the fix: `docker compose down` properly detaches the volume, then `docker volume rm` succeeds.

**Suggestion:**
```make
# Before:
@docker compose stop nextcloud 2>/dev/null || true
# After:
@docker compose down nextcloud 2>/dev/null || true
```

### 3. `info.xml` `max-version` needs bumping for NC 35

**File:** `eurooffice-nextcloud/appinfo/info.xml:38`

I had to bump `max-version` from `34` to `35` locally to make `make next` work with NC 35.0.0 dev. Without this change, the eurooffice app cannot install on NC 35 and `make next` fails with:

```
App "Nextcloud Office" cannot be installed because it is not compatible with this version of the server.
```

This will need to be committed as part of this PR (or a follow-up) to keep `make next` working as NC 35 ships.

## Suggestions for Improvement

### A. Naming consistency

The `local` / `mobile` and `next` / `next-mobile` naming is clear but slightly inconsistent. Consider:

- `make local` → `make desktop` (more explicit about what it targets)
- `make mobile` → `make lan` (more accurate — it's LAN mode, not just mobile)
- `make next` → `make next-nc` (avoids collision with git `next` branch)
- `make next-mobile` → `make next-lan` (consistent with above)

Alternatively, keep current names — they're concise and the README explains them well.

### B. `make pull` target is misleading

**File:** `Makefile:91`

```make
pull:
    docker compose pull && docker compose up -d && docker compose exec eo bash || true
```

This pulls all images (including `nextcloud:latest` and `onlyoffice/documentserver:latest`) and starts everything. But it doesn't run `refresh-urls` or print the banner. If you use it to update images, the URLs won't be configured.

**Suggestion:** Either add `refresh-urls` to the `pull` target, or rename it to `pull-and-start` or remove it entirely (users can run `docker compose pull && make` manually).

### C. `make build` path assumption

**File:** `Makefile:110-113`

```make
build:
    (cd ../build && docker buildx bake develop) && ...
```

This assumes the working directory is always `develop/` and `build/` is a sibling. Fine for documented usage, but if someone runs `make build` from elsewhere, it fails silently.

**Suggestion:** Add a guard at the top: `$(if $(findstring develop,$(CURDIR)),,$(error Run from DocumentServer/develop/))`

### D. `NC_BRANCH` volume naming collision

**File:** `docker-compose.next.yml:39`

```yaml
name: nc_data_${NC_BRANCH:-master}
```

If `NC_BRANCH` contains slashes (e.g. `feature/something`), Docker volume names will be invalid.

**Suggestion:** Sanitize: `name: nc_data_$(shell echo ${NC_BRANCH:-master} | tr '/' '_')`

### E. Consider a `make stop` target

Users may want to stop the stack without wiping data. Currently they need `docker compose down`. A simple `stop: docker compose down` would be convenient.

### F. Android emulator note in banner

**File:** `Makefile:38` and `README.md:49`

The `make mobile` banner shows the LAN IP (e.g. `192.168.178.194`), but the Android emulator needs `10.0.2.2` to reach the host. The `NEXTCLOUD_TRUSTED_DOMAINS` env var handles this at container creation, but the banner output is misleading.

**Suggestion:** Add a note in the `make mobile` banner: "For Android emulator, use `10.0.2.2` instead of the LAN IP shown above."

## Verdict

**Ship-ready with minor fixes.** The core functionality is solid and well-tested. The blockers are:

1. **Add timeout to `refresh-urls`** — prevents hanging forever on first boot of `nextcloud-docker-dev`.
2. **Fix `wipe-nc` / `wipe-next` to use `docker compose down`** — prevents silent failure when the container is running.
3. **Commit the `max-version` bump in `info.xml`** — required for `make next` to work with NC 35.

The other items (A-F) are nice-to-have improvements that can follow in a subsequent PR.

moodyjmz:
> Do we even need “make local” anymore? I find it a bit confusing that we have NC_VERSION and NC_BRANCH. I think julius's image should be enough, right? Maybe I'm misunderstanding something. What do you think?
> 
> Other than that, it looks good.

I left `make local` as it is simple. You only need changing IPs when working on mobile, otherwise it is just `make local` and stuff works for most usage.

moodyjmz:
> The `make mobile` banner shows the LAN IP (e.g. `192.168.178.194`), but the Android emulator needs `10.0.2.2` to reach the host. The `NEXTCLOUD_TRUSTED_DOMAINS` env var handles this at container creation, but the banner output is misleading.

This is from the readme :) However 10.0.2.2 is not going to work reliably, hence taking another approach. Two things need to happen, the machine IP needs to be used, and added to

moodyjmz:
I also realise, make next, whilst nice in idea, doesn't work in practice, as latest will be in a branch and needs loading that way.

PR #112, opened 2026-05-26
Feat: Rebrand, allow for custom brands

### LINKED ISSUE #106: Make package and image naming dynamic based on company name
As discussed this should include also paths in the packages and the about dialog references of the brand name

rikled:
These PRs replace some more onlyoffice logos, replace onlyoffice orange theme color by nextcloud blue theme color and add the ability to integrate custom brands.

depends on https://github.com/Euro-Office/server/pull/24 https://github.com/Euro-Office/document-server-integration/pull/9 https://github.com/Euro-Office/document-server-package/pull/9

### DIFF
--- a/build/.docker/bundle.bake.Dockerfile
+++ b/build/.docker/bundle.bake.Dockerfile
@@ -14,6 +14,11 @@
 
 FROM alpine AS bundle
     ARG BUILD_ROOT=/package
+    ARG PRODUCT_NAME_LOW
+    ARG COMPANY_NAME_LOW
+
+    ENV PRODUCT_NAME_LOW=${PRODUCT_NAME_LOW}
+    ENV COMPANY_NAME_LOW=${COMPANY_NAME_LOW}
     # --- Copy build files
     RUN mkdir -p /build/documentserver/sdkjs-plugins /build/documentserver/fonts \
                 /build/documentserver/server/FileConverter/lib /build/documentserver/server/tools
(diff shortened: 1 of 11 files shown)


chrip (approved):
Great work on the rebranding infrastructure. This lays a solid foundation for custom branding. Here are some observations:

**Positive:**
- Clean separation of branding variables (`COMPANY_NAME`, `COMPANY_NAME_LOW`, `PRODUCT_NAME`, `PRODUCT_NAME_LOW`) in `docker-bake.hcl`
- Using `regex_replace(lower(...))` for slugification is a good touch
- Supervisor configs now use environment variables instead of hardcoded paths, which is much more maintainable

**Suggestions:**

1. **`bundle.bake.Dockerfile` — `sed` substitutions on config JSON files** (lines 56-61, 80-85):
   Using `sed` to replace strings in JSON files at build time is fragile. If any config file contains the substring `euro-office` or `documentserver` in a value (not a key/path), it will be incorrectly replaced. Consider using `jq` or a more targeted replacement strategy for JSON files.

2. **`orchestrated.bake.Dockerfile` — hardcoded fallback value** (line 9):
   `ARG COMPANY_NAME_LOW=euro-office` hardcodes the fallback. This means builds that don't explicitly pass `COMPANY_NAME_LOW` will still produce `euro-office` branded images. Is this intentional? If the goal is full rebranding, the default should perhaps be empty or a generic placeholder.

3. **`standalone.bake.Dockerfile` — debconf selections** (lines 40-45):
   The debconf lines still reference `eurooffice` for the PostgreSQL user/db name. These should either be parameterized or documented as requiring manual override for custom brands.

4. **`entrypoint.sh` — string interpolation without quoting** (lines 4-5, 33, 44-46):
   Variables like `${EO_ROOT}-example` and `${EO_CONF}-example` are used without quotes. If paths contain spaces (possible with custom `COMPANY_NAME`), this could break. Recommend quoting: `"${EO_ROOT}-example"`.

5. **`ds-example.conf`** (lines 8-9):
   `%(ENV_EO_ROOT)s-example` and `%(ENV_EO_CONF)s-example` — using `-example` suffix assumes the example app always lives under the same root. This works for the current naming scheme but may be confusing for brands with different naming conventions. Consider an explicit `EXAMPLE_SUFFIX` variable.

6. **Missing newline at EOF** — `mariadb:$MARIADB_VERSION AS db-mariadb` section in `orchestrated.bake.Dockerfile` has no trailing newline (git diff shows `\ No newline at end of file`). Minor but worth fixing for cleanliness.

**Overall:** Solid foundation for the rebranding effort. Once the JSON `sed` fragility and unquoted variable usage are addressed, this should be ready to merge.

Issue #95, opened 2026-05-13
Docker env variable feature parity


juliusknorr:
Our entry point of the docker file is rather minimalistic, we should extend it to be on par with the upstream one: https://github.com/ONLYOFFICE/Docker-DocumentServer/blob/77ce899978efa62388cb79c59d8297bff740fce1/run-document-server.sh

Other related bug
- [x] #94

PR #116, opened 2026-05-28
feat(docker): standalone entrypoint env-var parity with upstream

### LINKED ISSUE #94: JWT_HEADER env variable is only partially working
@Rello mentioned in chat that the following json values are lacking if a custom JWT_HEADER=AuthorizationJwt is set.

```
"inbox": {
"header": "AuthorizationJwt",
"inBody": false
},
"outbox": {
"header": "AuthorizationJwt",
"inBody": false
}
```

Reference: https://github.com/ONLYOFFICE/Docker-DocumentServer/blob/77ce899978efa62388cb79c59d8297bff740fce1/run-document-server.sh#L397-L398

juliusknorr:
## Summary
- Brings `build/scripts/standalone/entrypoint.sh` to feature parity with the upstream [`run-document-server.sh`](https://github.com/ONLYOFFICE/Docker-DocumentServer/blob/77ce899978efa62388cb79c59d8297bff740fce1/run-document-server.sh), turning the standalone container into a drop-in replacement for the upstream one.
- Wires every relevant env var into `/etc/euro-office/documentserver/local.json` (and into nginx for SSL) before handing off to supervisord.
- Adds `netcat-openbsd`, `xxd`, and `openssl` to the standalone image; flips `ds-metrics` supervisor program to opt-in via `METRICS_ENABLED`.

## Env vars now supported
| Group | Vars |
| --- | --- |
| JWT | `JWT_ENABLED`, `JWT_SECRET`, `JWT_HEADER`, `JWT_IN_BODY`, and `*_INBOX` / `*_OUTBOX` variants for each |
| Database (postgres) | `DB_TYPE` (postgres only), `DB_HOST`, `DB_PORT`, `DB_NAME`, `DB_USER`, `DB_PWD` (alias: `DB_PASSWORD`, logs deprecation) |
| Redis | `REDIS_SERVER_HOST`, `REDIS_SERVER_PORT`, `REDIS_SERVER_USER`, `REDIS_SERVER_PASS`, `REDIS_SERVER_DB` |
| AMQP | `AMQP_URI`, `AMQP_HOST`, `AMQP_PORT`, `AMQP_USER`, `AMQP_PWD`, `AMQP_VHOST` |
| WOPI | `WOPI_ENABLED` (auto-generates RSA keypair under `/var/www/euro-office/Data`) |
| SSL/nginx | `SSL_CERTIFICATE_PATH`, `SSL_KEY_PATH`, `SSL_DHPARAM_PATH`, `SSL_VERIFY_CLIENT`, `ONLYOFFICE_HTTPS_HSTS_ENABLED`, `ONLYOFFICE_HTTPS_HSTS_MAXAGE`, `NGINX_WORKER_PROCESSES`, `NGINX_WORKER_CONNECTIONS`, `NGINX_ACCESS_LOG` |
| Metrics | `METRICS_ENABLED`, `METRICS_HOST`, `METRICS_PORT`, `METRICS_PREFIX` |
| Misc | `DS_LOG_LEVEL`, `PLUGINS_ENABLED`, `GENERATE_FONTS`, `USE_UNAUTHORIZED_STORAGE`, `ALLOW_PRIVATE_IP_ADDRESS`, `ALLOW_META_IP_ADDRESS`, `SECURE_LINK_SECRET` |

Defaults match upstream where reasonable (`JWT_ENABLED=true`, `WOPI_ENABLED=false`, `PLUGINS_ENABLED=true`, `METRICS_ENABLED=false`, `GENERATE_FONTS=true`, `JWT_IN_BODY=false`, `JWT_HEADER=Authorization`). `METRICS_PREFIX` defaults to `ds.` to line up with our cluster image instead of upstream's `.ds` (per review feedback).

## Behavior notes
- **Persisted secrets**: `JWT_SECRET`, `SECURE_LINK_SECRET`, and the WOPI keypair are auto-generated on first boot and stored under `/var/www/euro-office/Data/.private/` (directory 0700, files 0600 — including the WOPI RSA private key). Mount that directory as a volume to keep them stable across restarts; the JWT message printed at boot reminds operators of this.
- **Conditional bundled services**: `postgresql`, `redis-server`, and `rabbitmq-server` only start when the corresponding `*_HOST` points at `localhost`. For remote hosts the entrypoint waits for the port to open before starting supervisord.
- **DB type restriction**: the standalone image only ships `postgresql-client`, so non-postgres `DB_TYPE` values are rejected with a clear error. Users needing MariaDB/MySQL/MSSQL/Oracle should use the cluster image.
- **AMQP_VHOST normalization**: a value like `myvhost` is normalized to `/myvhost` before assembling the amqp:// URI, so users don't have to know the leading-slash detail.
- **Deliberately not implemented**: Let's Encrypt automation (`LETS_ENCRYPT_*`) and the upstream `ONLYOFFICE_DATA_CONTAINER` split-deployment mode. The cluster image covers split deployments; users with public TLS termination typically run a reverse proxy in front.
- **Behavior change**: boolean env vars (`ALLOW_PRIVATE_IP_ADDRESS`, `USE_UNAUTHORIZED_STORAGE`, etc.) are now compared against `"true"` rather than emptiness, matching upstream semantics. Anyone passing `1` or `yes` before will need to switch to `true`.

This PR also picks up the existing `JWT_HEADER` inbox/outbox fix (commit `5ba0cad`), so it closes both #94 and #95.

## Test plan

All items verified against a derivative image (nightly base + new entrypoint + new packages). The container-level checks were exercised end-to-end with live `docker run`s.

- [x] `JWT_SECRET=mysecret JWT_HEADER=AuthorizationJwt` → `secret.{browser,session,inbox,outbox}.string == "mysecret"` and `token.{inbox,outbox}.header == "AuthorizationJwt"` (closes #94)
- [x] `DB_PASSWORD=oldpwd` emits `WARNING: DB_PASSWORD is deprecated, use DB_PWD instead` on stderr and populates `services.CoAuthoring.sql.dbPass`
- [x] `DB_HOST=db.internal DB_PWD=… DB_USER=… DB_NAME=…` wires the external host into `services.CoAuthoring.sql`
- [x] `REDIS_SERVER_HOST=redis.internal REDIS_SERVER_PASS=… REDIS_SERVER_DB=2` populates `services.CoAuthoring.redis.{host,options.password,options.database}`
- [x] `AMQP_URI=amqp://u:p@rabbit.internal:5672/vhost` lands in `rabbitmq.url`; `AMQP_VHOST=myvhost` normalizes to `/myvhost`
- [x] `DB_TYPE=mysql` is rejected at startup with a clear error
- [x] `WOPI_ENABLED=true` generates the RSA keypair under `/var/www/euro-office/Data` and re-uses it across restarts (modulus stable); private key is `0600`
- [x] Auto-generated `JWT_SECRET` persists across restarts when the data dir is mounted
- [x] Baseline `docker run` prints the generated JWT secret on stdout and serves `/welcome/` (welcome page contains "Euro-Office Docs Community Edition")
- [x] With `DB_HOST=<external pg>`, `pgrep postgres` returns nothing inside the container and docservice's queries hit the external host (proven by schema-not-bootstrapped errors against the external DB)
- [x] With `REDIS_SERVER_HOST=<external redis>`, `pgrep redis-server` returns nothing inside the container; `redis.host` set to external host in `local.json`
- [x] `SSL_CERTIFICATE_PATH` + `SSL_KEY_PATH` produce a working HTTPS listener (`HTTP/2 200`) with HSTS header (`strict-transport-security: max-age=31536000`)
- [ ] Full `docker buildx bake standalone` from scratch — not re-verified in this round; the prior bake hung partway in the WASM/closure-compiler stages (likely OOM in the buildkit VM, not in our diff). Recommend re-running in CI or with more memory before merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/build/.docker/standalone.bake.Dockerfile
+++ b/build/.docker/standalone.bake.Dockerfile
@@ -29,7 +29,8 @@ ENV PRODUCT_NAME_LOW=${PRODUCT_NAME_LOW}
 RUN apt-get -y update && \
     ACCEPT_EULA=Y apt-get -yq install \
         postgresql postgresql-client redis-server rabbitmq-server \
-        nginx sudo gdb nginx-extras supervisor jq util-linux && \
+        nginx sudo gdb nginx-extras supervisor jq util-linux \
+        netcat-openbsd xxd openssl && \
     rm -rf /var/lib/apt/lists/*
 
 # Create the 'ds' user that is required by OnlyOffice scripts
--- a/build/configs/standalone/supervisor/ds-metrics.conf
+++ b/build/configs/standalone/supervisor/ds-metrics.conf
@@ -9,5 +9,5 @@ stdout_logfile_maxbytes=0
 stderr_logfile=%(ENV_EO_LOG)s/metrics/err.log
 stderr_logfile_backups=0
 stderr_logfile_maxbytes=0
-autostart=true
+autostart=false
 autorestart=true
--- a/build/scripts/standalone/entrypoint.sh
+++ b/build/scripts/standalone/entrypoint.sh
@@ -1,111 +1,358 @@
 #!/bin/sh
+set -e
 
-update_welcome_page() {
-    WELCOME_PAGE="${EO_ROOT}-example/welcome/docker.html"
-    EXAMPLE_DISABLED_PAGE="${EO_ROOT}-example/welcome/example-disabled.html"
+# --------------------------------------------------------------------
+# Paths
+# --------------------------------------------------------------------
+EO_ROOT="${EO_ROOT:-/var/www/euro-office/documentserver}"
+EO_LOG="${EO_LOG:-/var/log/euro-office/documentserver}"
+EO_CONF="${EO_CONF:-/etc/euro-office/documentserver}"
+DATA_DIR="/var/www/euro-office/Data"
+PRIVATE_DIR="${DATA_DIR}/.private"
+CONFIG_FILE="${EO_CONF}/local.json"
+LOG4JS_CONFIG="${EO_CONF}/log4js/production.json"
+EXAMPLE_CONF_DIR="/etc/euro-office/documentserver-example"
+EXAMPLE_LOCAL="${EXAMPLE_CONF_DIR}/local.json"
+NGINX_CONFIG_PATH="/etc/nginx/nginx.conf"
+NGINX_DS_DIR="${EO_CONF}/nginx"
+NGINX_DS_CONF="${NGINX_DS_DIR}/ds.conf"
+NGINX_DS_SSL_TMPL="${NGINX_DS_DIR}/ds-ssl.conf.tmpl"
+SUPERVISOR_CONF_DIR="/etc/supervisor/conf.d"
 
-    # Replace systemctl placeholder (set at build time) with docker+supervisorctl equivalent
-    sed -i 's|sudo systemctl start ds-example|sudo docker exec $(sudo docker ps -q) supervisorctl start ds:example|g' \
-        "$EXAMPLE_DISABLED_PAGE"
+# --------------------------------------------------------------------
+# Defaults (aligned with upstream run-document-server.sh)
+# --------------------------------------------------------------------
+JWT_ENABLED="${JWT_ENABLED:-true}"
+JWT_HEADER="${JWT_HEADER:-Authorization}"
+JWT_IN_BODY="${JWT_IN_BODY:-false}"
 
-    if [ -e "$WELCOME_PAGE" ]; then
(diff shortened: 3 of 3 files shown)


juliusknorr:
Addressed Copilot's three review comments in ed75701:

- `METRICS_PREFIX` default changed from `.ds` (upstream's odd convention) to `ds.` (matches our cluster entrypoint, also the more typical statsd prefix shape).
- `AMQP_VHOST` normalized to add a leading slash if missing, so `AMQP_VHOST=myvhost` produces a valid URI.
- WOPI private key gets `chmod 600` after generation so the RSA key isn't world-readable on the Data volume.

rikled:
tested locally, works fine. fixed some missed static paths.

rikled (approved):

PR #137, opened 2026-06-09
feat: enhance dockerfiles

RetGal:
Simplify and unify the Dockerfiles. 
This should also result in slightly smaller image sizes.

### DIFF
--- a/build/.docker/develop.bake.Dockerfile
+++ b/build/.docker/develop.bake.Dockerfile
@@ -9,18 +9,20 @@
 
 FROM finalubuntu AS develop
     ARG PRODUCT_VERSION
+    # Avoid interactive prompts during package install
+    ARG DEBIAN_FRONTEND=noninteractive
     ENV TZ=Etc/UTC
 
     RUN apt-get update && \
         curl -fsSL https://deb.nodesource.com/setup_20.x | bash - && \
-        apt-get install -y nodejs openjdk-21-jdk wget zip brotli && \
+        apt-get install -yq --no-install-recommends nodejs openjdk-21-jdk wget zip brotli && \
         npm install -g @yao-pkg/pkg grunt-cli && \
         rm -rf /var/lib/apt/lists/*
 
     RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone && \
-        apt-get -y update && \
-        apt-get -y upgrade && \
-        apt-get -y install \
+        apt-get update && \
+        apt-get upgrade -y && \
+        apt-get install -yq --no-install-recommends \
                     git \
                     curl \
                     sudo \
--- a/build/.docker/packages.bake.Dockerfile
+++ b/build/.docker/packages.bake.Dockerfile
@@ -21,6 +21,8 @@ FROM ubuntu:24.04 AS package
     ARG BUNDLE_BASE="/build/bundle"
     ARG OUT_DIR="${BUNDLE_BASE}/${COMPANY_NAME_LOW}/documentserver"
     ARG EXAMPLE_OUT="${BUNDLE_BASE}/${COMPANY_NAME_LOW}/documentserver-example"
+    # Avoid interactive prompts during package install
+    ARG DEBIAN_FRONTEND=noninteractive
 
     ENV PRODUCT_VERSION=${PRODUCT_VERSION}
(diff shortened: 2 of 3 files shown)


chrip (changes requested):
Thanks a lot for this PR, @RetGal! These are solid Dockerfile improvements — moving `DEBIAN_FRONTEND` to `ENV`, adding `--no-install-recommends`, and cleaning up apt caches are all good practices. A few things to look at before we merge:

**File-level comments:**

1. `build/.docker/develop.bake.Dockerfile` — `curl` was removed from the install list. Is this intentional? `curl` is used at build time for the nodesource setup (already available from the base image), but it was also kept as a persistent tool in the develop image for developers. If dev workflows or scripts rely on `curl` being present, this will break them. If it's safe to remove, just worth documenting the intent.

2. `build/.docker/develop.bake.Dockerfile` — The original had `rm -rf /var/lib/apt/lists/*` at the end of the RUN. The combined version dropped it, so the apt cache will persist in the image. Adding it back would actually help with the image size goal mentioned in the PR description.

3. `build/.docker/standalone.bake.Dockerfile` — The `rm -rf /var/lib/apt/lists/* /tmp/*.deb` cleanup added here is a nice touch, and removing the redundant `apt-get update` before the .deb install is correct since packages are already cached from the prior RUN.

4. `build/.docker/develop.bake.Dockerfile` — Combining the two RUN statements into one removes a Docker cache breakpoint. That means any change to the system tools (git, sudo, etc.) will force a rebuild of the expensive nodejs + npm global installs. Worth considering whether splitting them back into two RUNs would help with iterative build times.

Overall really appreciated — small nitpicks on top of otherwise clean improvements.

Assisted-by: OpenCode:Qwen3.6-27B

RetGal:
@chrip Thanks for your review!
I re-added the installation of `curl` and split the two RUN statements - as your argument about the Docker cache breakpoint is convincing! After that change, also having another separate `rm -rf /var/lib/apt/lists/*` made sense again =)

chrip (changes requested):
Thanks for the quick turnaround, @RetGal — confirming my earlier round is fully addressed in `e775a0f`:

- ✅ `curl` re-added to the `develop` install list
- ✅ `rm -rf /var/lib/apt/lists/*` restored on both RUNs
- ✅ The two RUNs split back apart, so the node/npm layer keeps its cache breakpoint

All good there. Two smaller things I noticed on a closer pass:

**1. `DEBIAN_FRONTEND=noninteractive` as `ENV` leaks into the final runtime image**
I flagged moving it to `ENV` as a positive last time, but I want to walk that back for `standalone` specifically: `finalubuntu` is the *shipped* image, so `ENV DEBIAN_FRONTEND=noninteractive` persists a permanent noninteractive debconf frontend into every running container — the documented Docker caveat (it can surprise anyone running `apt`/`dpkg` interactively inside the container later). A build-scoped `ARG` gets you the same build-time behavior without the leak:

```dockerfile
ARG DEBIAN_FRONTEND=noninteractive
```

It's still visible to all build `RUN` steps but doesn't survive into the image. Same reasoning applies to `develop` (a dev image people shell into); `packages` is a pure build stage so it doesn't matter there.

**2. `standalone` `/tmp` cleanup leaves the `.rpm` artifacts behind**
The `packages` scratch stage carries *both* `.deb` and `.rpm` outputs, and `COPY --from=packages / /tmp/` pulls all of them in. The new cleanup only removes `*.deb`:

```dockerfile
rm -rf /var/lib/apt/lists/* /tmp/*.deb
```

so the `.rpm` files stay in the final image, partly undercutting the size goal. Since nothing else in `/tmp` is needed past this point:

```dockerfile
rm -rf /var/lib/apt/lists/* /tmp/*
```

(or list `/tmp/*.deb /tmp/*.rpm` explicitly).

Nice cleanup overall — happy to approve once these two are settled (or if you'd rather keep `ENV`, a one-line note on the intent is fine by me).

Assisted-by: ClaudeCode:Opus-4.8

RetGal:
@chrip Done as suggested - I wasn't aware that the ENVs leak into the runtime image.

chrip (approved):
All prior reviewer feedback has been fully addressed.

chrip:
**built & measured on arm64:** 
`standalone` image 4.49 GB → 4.31 GB (−180 MB / −4.0%)
`develop` image 7.72 → 7.41 GB (−315 MB / −4.1%)

PR #138, opened 2026-06-09
fix: properly tag final release builds

### LINKED ISSUE #309: v9.3.2 release assets still report Version: 9.3.1-dev.1 (same as #247)
## Description

The issue reported in #247 (packages published as v9.3.2 report an
incorrect version string "9.3.1-dev.1") is still present in the final
stable v9.3.2 release (June 25, 2026), despite this release's
changelog including changes to the tagging script ("Tag Release
Script: Fixed release tagging logic - corrected build number
initialization and calculation").

## Steps to reproduce

1. Download the asset from the v9.3.2 release:
   https://github.com/Euro-Office/DocumentServer/releases/download/v9.3.2/euro-office-documentserver_9.3.1-dev.1_amd64.deb

   Note that the filename itself still contains "9.3.1-dev.1"
   instead of "9.3.2".

2. Check the package's internal metadata:
   dpkg -I euro-office-documentserver_9.3.1-dev.1_amd64.deb | grep -i version

## Expected behavior

Version: 9.3.2

## Actual behavior

Version: 9.3.1-dev.1

## Additional notes

All six assets in the v9.3.2 release (deb amd64/arm64, rpm
x86_64/aarch64) carry the same incorrect "9.3.1-dev.1" naming.
The tagging-script fix described in the changelog does not appear to
have propagated to the build/packaging pipeline that generates the
actual version number embedded in the packages.

Claudio

rikled:
The tag_release.sh script was unable to create non-pre-release tags. This fix makes the --pre-id parameter optional. So if no pre-id is given it only adds the build number.

Example: Version: v9.3.1, build 5

- if "--pre-id tp", then it creates v9.3.1-tp.5
- if "--pre-id" is not provided, then it creates v9.3.1-5

### DIFF
--- a/scripts/tag_release.sh
+++ b/scripts/tag_release.sh
@@ -27,7 +27,7 @@ PUSH=0
 usage() {
     echo "Usage: $0 [-b BUILD_NUMBER] [-p PRE_ID] [--dry-run] [--push]"
     echo ""
-    echo "Tags the main repo and all submodules with vVERSION[-PRE_ID].BUILD_NUMBER"
+    echo "Tags the main repo and all submodules with vVERSION[-PRE_ID]+BUILD_NUMBER"
     echo ""
     echo "Options:"
     echo "  -b, --build BUILD_NUMBER   Build number (default: last build + 1)"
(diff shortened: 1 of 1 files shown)


DmySyz:
Should we have an option to create tags without build number as well? Just the version.

rikled:
To me, it makes more sense to always include a build number (especially after today, when we really just had to rebuild to fix a problem). Does anyone else have any thoughts on this? @juliusknorr ?

juliusknorr:
Yes, I agree a build number helps to avoid version bumps if we actually just bump dependency packages or build tooling.

scripts/tag_release.sh
```diff
@@ -66,10 +66,19 @@ while [[ $# -gt 0 ]]; do
     esac
 done
 
+# Build the tag prefix. With a pre-release id it's "vX.Y.Z-ID.", without it "vX.Y.Z-".
+if [ -n "$PRE_ID" ]; then
+    TAG_PREFIX="v${VERSION}-${PRE_ID}."
+else
+    TAG_PREFIX="v${VERSION}-"
```
juliusknorr: If we want to stick to semantic versioning the build number would be separated by 

https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions

```
E.g. with build id 4
9.3.1+4
9.3.1-beta.1+4
```

Issue #177, opened 2026-06-13
Bug: WOPI broken: entrypoint.sh writes wopi.privateKey with literal \n instead of real newlines → DECODER routines::unsupported


filerun:
AI generated post, but the fix works for me... so, I hope it is useful to you too.

## Summary

With `WOPI_ENABLED=true`, opening any document in WOPI mode fails. The editor shows the generic *"Sorry, editor could not be loaded. Please contact your administrator."* The DocService log shows proof-signing throwing on every `checkFileInfo`:

```
nodeJS - wopi checkFileInfo error status=NaN (Unknown error):
Error: error:1E08010C:DECODER routines::unsupported
    at signOneShot (node:internal/crypto/sig:169:15)
    at generateProofSign (/snapshot/server/DocService/sources/wopiUtils.js)
    at Object.fillStandardHeaders (/snapshot/server/DocService/sources/wopiUtils.js)
    at checkFileInfo (/snapshot/server/DocService/sources/wopiClient.js)
```

## Root cause

In `entrypoint.sh`, the WOPI private key is flattened before being injected into `local.json`:

```sh
WOPI_PRIVATE_KEY_DATA=$(awk '{printf "%s\\n", $0}' "$WOPI_PRIVATE_KEY")
```

This joins the PEM lines with the **two literal characters** `\` + `n`, then passes the value to `jq --arg wopiPriv "$WOPI_PRIVATE_KEY_DATA"`.

`jq --arg` treats its value as a **literal string** and does not interpret `\n` as a newline, so jq JSON-encodes the backslash-n verbatim. The resulting `local.json` contains:

```json
"privateKey": "-----BEGIN PRIVATE KEY-----\nMIIEvAIB...==\n-----END PRIVATE KEY-----\n"
```

…where `\n` is literal text, **not** an encoded newline. When DocService parses the JSON it gets a PEM string with zero real line breaks, which OpenSSL cannot decode — hence `DECODER routines::unsupported` inside `generateProofSign`.

Only the private key is affected; `publicKey`/`modulus`/`exponent` go through base64/numeric paths and are fine. The generated keypair itself is valid — restoring the newlines makes it parse and sign normally.

## Reproduction

1. Run the image with `WOPI_ENABLED=true`.
2. Inspect the generated config:
   ```sh
   docker exec <ctr> python3 -c 'import json; \
     k=json.load(open("/etc/euro-office/documentserver/local.json"))["wopi"]["privateKey"]; \
[...]

PR #182, opened 2026-06-14
Fix WOPI proof key signing failure

### LINKED ISSUE #127: WOPI server fails to start because of wrong encoded private key
WOPI Server fails to start because of this line in the docker entrypoint.sh

https://github.com/Euro-Office/DocumentServer/blob/863832a5e854de631ecd55bc1d94e41c641627ff/build/scripts/standalone/entrypoint.sh#L146

Replacing it with `cat` resolve the issue.
```diff
-  WOPI_PRIVATE_KEY_DATA=$(awk '{printf "%s\\n", $0}' "$WOPI_PRIVATE_KEY")
+  WOPI_PRIVATE_KEY_DATA=$(cat "$WOPI_PRIVATE_KEY")
```

Docserver error
```
[2026-06-05T06:03:15.243] [WARN] [localhost] [docId] [userId] nodeJS - Express server starting...
[2026-06-05T06:03:15.247] [WARN] [localhost] [docId] [userId] nodeJS - notifyLicenseExpiration(): expiration date is not defined
[2026-06-05T06:03:15.247] [WARN] [localhost] [docId] [userId] nodeJS - notifyLicenseExpiration(): expiration date is not defined
[2026-06-05T06:03:15.362] [WARN] [localhost] [docId] [userId] nodeJS - Express server listening on port 8000 in production-linux mode. Version: 9.3.1. Build: 37
[2026-06-05T06:04:28.412] [ERROR] [localhost] [48cadb6dd7750c7fe0148647427489d53c7075f7f9652b7c23cae8ecac40f26d] [userId] [bb91608a-4925-419f-9ac2-4f860322070c] nodeJS - wopi checkFileInfo error status=NaN (Unknown error):Error: error:1E08010C:DECODER routines::unsupported
    at signOneShot (node:internal/crypto/sig:169:15)
    at node:internal/util:441:7
    at new Promise (<anonymous>)
    at signOneShot (node:internal/util:427:12)
    at generateProofSign (/snapshot/server/DocService/sources/wopiUtils.js)
    at Object.fillStandardHeaders (/snapshot/server/DocService/sources/wopiUtils.js)
    at checkFileInfo (/snapshot/server/DocService/sources/wopiClient.js)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

I can open a PR if you like!

delmarguillen:
## Summary

The `awk` command on line 146 of `entrypoint.sh` flattens the PEM private key by replacing real newlines with literal `\n` characters:

```sh
WOPI_PRIVATE_KEY_DATA=$(awk '{printf %s\\n, $0}' $WOPI_PRIVATE_KEY)
```

When `jq --arg` receives this, it treats `\n` as literal text and escapes it again. The resulting key in `local.json` has no real newlines, which OpenSSL cannot decode:

```
error:1E08010C:DECODER routines::unsupported
    at signOneShot (node:internal/crypto/sig:169:15)
    at generateProofSign (/snapshot/server/DocService/sources/wopiUtils.js)
```

This breaks any WOPI integration that requires proof key verification. Tested on both x86_64 and ARM64.

## Fix

Replace `awk` with `cat` so `jq --arg` receives the raw PEM with real newlines and encodes them correctly in JSON.

One line change. Credit to @tiran133 for identifying the fix in #127.

## How I found this

I built an integration of Euro-Office with ownCloud OCIS ([repo](https://github.com/delmarguillen/euro-office-ocis)). The WOPI `checkFileInfo` call failed on every document open.

## Tested on

- DocumentServer 9.3.1 Build 37 (`ghcr.io/euro-office/documentserver:latest`)
- VPS: x86_64, Ubuntu 24.04, Docker 29.5.3
- Local: ARM64 (Surface Pro 9, Podman 5.8.3, WSL2)

### DIFF
--- a/build/scripts/standalone/entrypoint.sh
+++ b/build/scripts/standalone/entrypoint.sh
@@ -143,7 +143,7 @@ if [ "$WOPI_ENABLED" = "true" ]; then
   fi
   WOPI_MODULUS=$(openssl rsa -pubin -inform "MS PUBLICKEYBLOB" -modulus -noout -in "$WOPI_PUBLIC_KEY" | sed 's/Modulus=//' | xxd -r -p | openssl base64 -A)
   WOPI_EXPONENT=$(openssl rsa -pubin -inform "MS PUBLICKEYBLOB" -text -noout -in "$WOPI_PUBLIC_KEY" | grep -oP '(?<=Exponent: )\d+')
-  WOPI_PRIVATE_KEY_DATA=$(awk '{printf "%s\\n", $0}' "$WOPI_PRIVATE_KEY")
+  WOPI_PRIVATE_KEY_DATA=$(cat "$WOPI_PRIVATE_KEY")
   WOPI_PUBLIC_KEY_DATA=$(openssl base64 -in "$WOPI_PUBLIC_KEY" -A)
 fi
 


delmarguillen:
This fixes #127, #160, and #177. The root cause is that `openssl genpkey` produces a PKCS#8 key that Node.js `crypto.sign()` cannot use with the bundled OpenSSL 3.x backend. Switching to `openssl genrsa -traditional` generates a PKCS#1 key that works.

streaminganger:
> The workaround was converting the key after generation, but the proper fix is generating it in the right format from the start.

this doesn't make sense to me. i would have thought the patch by @tiran133 was a proper fix instead. the existing algorithm has been in use since onlyoffice. the root problem being that jq args will treat \n literally and then escape it result in an invalid key. @filerun analysis in #177 is also correct.

delmarguillen:
Thanks for pointing that out. I updated the PR with the proper fix from #127, the actual issue is the awk line double-escaping newlines before jq handles them.

delmarguillen:
Tested the updated fix on a VPS (x86_64, Ubuntu 24.04) and it works correctly.

juliusknorr (approved):

PR #186, opened 2026-06-15
feat(develop): parallel test servers via eo.sh

juliusknorr:
Adds `develop/eo.sh` to run multiple isolated DocumentServer test instances in parallel from the prebuilt multi-arch `:latest-dev` image, no full docker build required. Useful for quickly verifying submodule changes (incl. by coding agents) and for testing several branches/worktrees at once.

Each instance is a single self-contained container (its own Postgres/Redis/RabbitMQ/Nginx/supervisord) that mounts the current git working tree at `/develop`, so the existing in-container `make` targets rebuild local changes in place.

```
eo.sh up [--force] <name> [port]   # start; auto host port; waits for /healthcheck; prints URLs
eo.sh build <name> <make-target>   # web-apps-dev | sdkjs | server/docservice | core/x2t | …
eo.sh exec <name> [cmd…]           # shell/command inside (TTY-aware)
eo.sh logs <name>                  # follow logs
eo.sh ls                           # instances with status, port, URL, mount path
eo.sh down <name>… | --all         # stop & remove
```

Instances run with JWT enabled (shared dev secret), `EXAMPLE_ENABLED` and `WOPI_ENABLED`, so the bundled example app works and documents open. `up` prints create-new editor URLs for all document types.

Also corrects a stale `develop/README.md` note (a multi-arch `:latest-dev` incl. arm64 is published, so `make pull` works on Apple Silicon) and documents the new workflow.

## Verification
- Multiple instances run concurrently on distinct auto-assigned ports; healthchecks pass; example editor opens a document end-to-end.
- `eo.sh build … web-apps-dev` runs a real in-container build and reloads nginx.
- `bash -n` + shellcheck clean; kept bash 3.2 compatible (macOS system bash).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -45,6 +45,24 @@ Troubleshooting: if `make` fails with `Makefile: No such file or directory`, the
 - **Conversion check**: opening a `.docx`/`.xlsx` in the example app exercises FileConverter + x2t; watch `converter/out.log` for errors.
 - **Nextcloud connector flow** (only when testing the integration): `docker compose up -d`, then `make refresh-urls` to wait for install and wire URLs/JWT; Nextcloud at `http://localhost:8081/` (admin/admin).
 
+### Running instances (`eo.sh`)
+
+`develop/eo.sh` runs one or more test servers as self-contained containers from the same image — current tree mounted at `/develop`, auto-assigned port. Starts detached (agent-friendly, unlike `make local`):
+
+```sh
+cd develop
+./eo.sh up <name>                  # start; waits for /healthcheck; prints URL + per-type editor URLs
+./eo.sh build <name> web-apps-dev  # in-container make targets (sdkjs, core/x2t, server/docservice, …)
+./eo.sh exec <name> <cmd…>         # run inside (TTY-aware)
+./eo.sh ls | logs <name> | down <name>|--all   # manage; `up --force` to recreate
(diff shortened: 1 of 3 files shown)


chrip (approved):
Solid tool — well-designed bash (3.2 compatible, shellcheck-clean), solves a real need. JWT secret, image, and Makefile mount all match `docker-compose.yml`.

One question: `EXAMPLE_ENABLED` and `WOPI_ENABLED` are set to `true` in `eo.sh` instances but are **not** set in the existing `docker-compose.yml`. Is this intentional? The compose stack relies on Nextcloud's built-in example integration for those, while standalone `eo.sh` instances need them explicit to make the bundled example app work. Just wanted to confirm the design choice is deliberate.

Otherwise, LGTM.

juliusknorr:
> EXAMPLE_ENABLED and WOPI_ENABLED are set to true in eo.sh instances but are not set in the existing docker-compose.yml. Is this intentional? The compose stack relies on Nextcloud's built-in example integration for those, while standalone eo.sh instances need them explicit to make the bundled example app work. Just wanted to confirm the design choice is deliberate.

Depends on what you want to use the compose setup for. It could be useful to extend the compose file with that indeed, to be able to work on wopi integration with Nextcloud or use the example app there. Would say something that can be added if needed there.

PR #197, opened 2026-06-16
test(e2e): add editor smoke tests for typing, formulas and slides

juliusknorr:
## Summary

Adds Playwright smoke tests exercising basic editing in each editor, run against the example app:

- **Document** — type text, apply bold/italic, verify the text round-trips via the automation API
- **Spreadsheet** — enter numbers and a `=A1+A2` formula, verify the formula is accepted in the cell
- **Presentation** — build a three-slide deck and verify the slide count

A shared `helpers.ts` opens a new document from the example page, waits for the editor iframe to finish loading, and exposes `editorApi()` to call `window.Asc.editor` inside the iframe.

## Notes

- Element ids are renamed in this build, so tests drive the editors via keyboard shortcuts and the `asc_*` automation API rather than clicking toolbar buttons.
- Spreadsheet test asserts the formula is *accepted*, not that it computes `15`: `asc_getCellInfo().asc_getText()` returns the formula string and `GetSelectedText()` is empty for cell selections. A TODO documents the clipboard-based computed-value check (needs clipboard permissions in the project config).
- Presentation slides are added through `AddSlide()` because `Ctrl+M` only fires when the thumbnail panel is focused.

## Testing

Verified locally against a running example server:

```
3 passed (11.7s)
```

## AI assistance

Tests authored with AI assistance (Claude Code, claude-opus-4-8); reviewed and verified by the contributor.


Alex-Arsys:
LGTM

e2e/tests/document-editing.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+import { openNewEditor, editorApi } from './helpers';
+
+test.describe('Document editor - typing and formatting', () => {
+  test('type text and apply bold/italic', async ({ page }) => {
+    const { editorPage, frame } = await openNewEditor(page, 'a.try-editor.word', /\.docx/);
+
+    await editorPage.keyboard.type('Hello world');
+    await editorPage.keyboard.press('Control+A');
+    await editorPage.keyboard.press('Control+b');
+    await editorPage.keyboard.press('Control+i');
```
j-base64, line 11: commented these lines and the test still passed.
maybe we can test these two bold/italic actions with 
```
const isBold = await editorApi(editorPage, (api) => api.get_TextProps().get_TextPr().get_Bold());
const isItalic = await editorApi(editorPage, (api) => api.get_TextProps().get_TextPr().get_Italic());
expect(isBold).toBe(true);
expect(isItalic).toBe(true);
```
Tested it locally, fails commenting line 10-11 and passes with them uncommented.

j-base64 (commented):
Looks good, 

Just one tiny comment inline about bold/italic test.

also, not sure of importance or root cause,  noticed that running with --workers 1 all tests pass clean. 
with parallel workers, tests intermittently open the wrong editor/extension

moodyjmz:
**TL;DR:** The spreadsheet cell reads (`spreadsheet-formula.spec.ts:32-40`) race the selection change and can flake. `nameBox.press('Enter')` resolves when the key is *dispatched*, not when the editor finishes moving the selection and `asc_getCellInfo()` reflects the new cell — and `expect(await cellText(...)).toBe(...)` is a one-shot with no retry. Wrap the reads in `expect.poll` (the presentation test already does this for the analogous async-count case):

```ts
await select('A1');
await expect.poll(() => cellText(editorPage)).toBe('5');
await select('A2');
await expect.poll(() => cellText(editorPage)).toBe('10');
await select('A3');
await expect.poll(() => cellText(editorPage)).toBe('=A1+A2');
```

`retries: 1` will mostly hide the flake rather than fix it, which is worse — hence the poll.

<details>
<summary>Detail</summary>

The spreadsheet runs calc off the UI thread, so there's a genuine async gap between the name-box navigation and `asc_getCellInfo()` returning the new cell's info. The tell is that `presentation-slides.spec.ts` already uses `await expect.poll(() => slideCount(...)).toBe(3)` for exactly this class of problem (a value that updates asynchronously after a mutation) — the same treatment just wasn't applied to the cell reads.

Reviewed with AI assistance (Claude Code, claude-opus-4-8); grounded against the PR diff and the existing `example-page.spec.ts` / `playwright.config.ts` on `main`.
</details>

chrip (changes requested):
## Verification
| Document test verifies bold/italic (PR desc: "apply bold/italic") | Only assertions are `#slot-btn-undo button` not `disabled` and `asc_GetSelectedText()` contains 'Hello world'. Plain typing enables undo and produces the text, so the `Ctrl+b`/`Ctrl+i` keystrokes (lines 10-11) are effectively unverified — matches @j-base64's local finding. | ❌ |
| Spreadsheet reads are deterministic | `select()` does `nameBox.press('Enter')` (resolves on key dispatch), then a one-shot `expect(await cellText(...)).toBe(...)`. `asc_getCellInfo()` can lag the selection move → flake. Presentation test uses `expect.poll` for the same async-value class; spreadsheet does not. | ❌ |
| Presentation slide count uses proper retry wait | Uses `await expect.poll(() => slideCount(editorPage)).toBe(3)` — correct, deterministic. | ✓ |
| Spreadsheet asserts formula *accepted*, not computed (=15) | Correct and honestly documented in the TODO; `asc_getText()` returns the formula string, computed-value path needs clipboard perms. | ✓ |
| No arbitrary `sleep`/`waitForTimeout` | Confirmed — waits are `expect`-with-timeout, `waitForURL`, `toBeHidden` on `#loading-mask`, `expect.poll`. Good. | ✓ |
| e2e CI job runs these specs and is green | `build.yml` has an `e2e` job (`needs: [build, manifest]`, self-hosted) that runs `npx playwright test` in `e2e/` — wiring exists. But on this PR's run the `build` jobs FAILED, so `e2e` was **SKIPPED**. The tests were never exercised in CI. | ❌ |
| Build failure is caused by this PR | No. Root cause is a 3rd-party core build step: `gclient_paths.patch failed … patch does not apply` / `Common/3dParty/build_3rdparty.py failed!` — unrelated to adding `.ts` test files. | ✓ (unrelated) |
| Commit hygiene (conventional, DCO, AI disclosure) | `test(e2e): …`, `Signed-off-by:` present (DCO check green), `Assisted-by: ClaudeCode:claude-opus-4-8` trailer, AI assistance disclosed in body. | ✓ |
| Diff scope | Exactly 4 new files under `e2e/tests/`; `package.json`/`playwright.config.ts`/`constants.ts`/`example-page.spec.ts` are pre-existing scaffold, not touched here. | ✓ |

## Issues & Suggestions

### 🔴 Blocking
- **Document test does not verify the formatting it advertises** — `e2e/tests/document-editing.spec.ts:10-11`. The `Ctrl+b`/`Ctrl+i` presses have no corresponding assertion; @j-base64 verified locally that commenting them out still passes. A "bold/italic" smoke test that stays green when formatting is entirely removed gives false confidence. Adopt @j-base64's concrete fix:
  ```ts
  const isBold = await editorApi(editorPage, (api) => api.get_TextProps().get_TextPr().get_Bold());
  const isItalic = await editorApi(editorPage, (api) => api.get_TextProps().get_TextPr().get_Italic());
  expect(isBold).toBe(true);
  expect(isItalic).toBe(true);
  ```
  (Or rename the test/PR to "type text" and drop the formatting claim — but verifying is clearly the intent.)

### ⚠️ Major
- **Spreadsheet cell reads can flake** — `e2e/tests/spreadsheet-formula.spec.ts` (the `select`→`expect(await cellText(...))` block, ~lines 28-37). The name-box `Enter` resolves before `asc_getCellInfo()` reflects the new cell. Wrap each read in `expect.poll(() => cellText(editorPage)).toBe(...)`, mirroring the presentation test. Note `retries: 1` in `playwright.config.ts` will *mask* rather than *fix* this — @moodyjmz's point is correct.
- **Parallel-worker cross-talk** — @j-base64 reports tests intermittently open the wrong editor with multiple workers (clean with `--workers 1`). `playwright.config.ts` sets `fullyParallel: false` but does not pin `workers: 1`, so spec files still run concurrently and appear to collide on the shared example app / new-tab targeting. Worth pinning `workers: 1` (the workflow comment already notes the e2e job assumes a single runner) or root-causing the tab mismatch before this lands, otherwise CI will be flaky.
- **No CI validation yet** — the `e2e` job was SKIPPED because `build` failed on an unrelated 3rd-party patch error. There is currently no green run proving these specs pass in CI (only the author's local "3 passed"). Recommend rebasing onto current `main` to get a clean build + an actual `e2e` run before merging.

### ℹ️ Minor / 💡 Suggestions
- **Presentation test drives the API, not the UI** — `presentation-slides.spec.ts` adds slides via `api.AddSlide()` because `Ctrl+M` needs thumbnail focus (documented). Reasonable for a smoke test, but it exercises the automation API more than user interaction; a follow-up focusing the thumbnail panel and using the shortcut would be closer to a real user path. 💡
- **Selector robustness** — tests rely on internal ONLYOFFICE ids (`#slot-btn-undo`, `#ce-cell-name`, `#editor_sdk`, `#loading-mask`). These can shift on upstream bumps, but this matches the existing `example-page.spec.ts` convention, so it's consistent, not new debt. ℹ️
- **`editorApi` serialization** — `helpers.ts` rebuilds the passed fn via `new Function` inside `frame.evaluate`. The "must be self-contained (no closures)" caveat is documented; fine for tests. ℹ️
- **SPDX headers** — the new `.ts` files carry no SPDX header, but the pre-existing `example-page.spec.ts` also has none, so the e2e dir convention is header-less. Non-blocking; flag only if the repo later standardizes. ℹ️

## Verdict
**Request changes** — Solid, well-scoped test scaffolding, but the headline document test provably does not verify bold/italic (already flagged inline by @j-base64), the spreadsheet reads have a genuine race (@moodyjmz), and no CI run has actually executed these specs (build failed on an unrelated 3rd-party patch, skipping the e2e job). Fix the bold/italic assertion, poll the spreadsheet reads, and get a green e2e run — after that this is an easy approve.

moodyjmz:
@chrip pushed two commits addressing both blocking points:

**Rebase (`bc684dc`)** — merged current `main` in, clean merge-tree, no conflicts. `mergeStateStatus` is `CLEAN` now, so this should actually get a real `e2e` run this time instead of bouncing off the `/core` build issue from July (that's been green on main for a while now).

**Bold/italic + spreadsheet race (`810c299`)** — both fixed as you and j-base64 flagged:
- `document-editing.spec.ts` only asserted "some edit happened" (undo enabled, text present), never that `Ctrl+b`/`Ctrl+i` actually applied formatting. Added `get_TextProps().get_TextPr().get_Bold()`/`get_Italic()` assertions per your suggestion — and to make sure they're not just as vacuous as the old ones, ran a negative-control pass with the keystrokes commented out first: confirmed it now correctly fails (`1 failed`) before restoring them and confirming green again.
- `spreadsheet-formula.spec.ts`'s cell reads raced `asc_getCellInfo()` against the selection update, same class of async-value issue the presentation test already guards against. Wrapped all three reads in `expect.poll()`, matching that existing pattern.

Both re-verified end-to-end against a live container (`--workers=1`, per the cross-talk issue) — all 3 specs pass, ~26s.

Not touched: the `workers: 1` config pin and the parallel cross-talk root cause — left those as the fast-follow you already flagged as non-blocking, rather than scope-creeping this PR further.

chrip (approved):
Both of my blocking points are properly fixed and I verified them against
sdkjs.

PR #200, opened 2026-06-17
chore: remove damengdb dependency

juliusknorr:
Removes the `onlyoffice/damengdb` Docker image dependency and the `disql` wrapper script, as the Dameng database support has been dropped.

Closes #165


rikled (approved):
lgtm

PR #216, opened 2026-06-18
fix(docker): give ds services a writable HOME so sharp can load

juliusknorr:
Set `HOME=/home/ds` for all `ds`-owned supervisor programs and create the directory in the standalone image.

## Why

`supervisord` runs as root and does not reset `HOME` when dropping to `user=ds`, so the node services inherited `HOME=/root` and could not write their cache. sharp/pkg failed to extract its native module to `/root/.cache` (`EACCES`), which disabled image processing (`processImageOptimal` returned the original buffer).

## Test

In a running standalone container, apply the same `HOME` env to `ds-docservice`/`ds-converter`, `mkdir -p /home/ds && chown ds:ds /home/ds`, restart the services, open a document containing images, and confirm the `Sharp module failed to load` / `processImageOptimal: Sharp module not available` warnings no longer appear in `docservice/out.log`.

---
AI disclosure: changes drafted with AI assistance (ClaudeCode, model claude-opus-4-8); see the `Assisted-by` commit trailer. Please review before merging.

### DIFF
--- a/build/.docker/standalone.bake.Dockerfile
+++ b/build/.docker/standalone.bake.Dockerfile
@@ -59,6 +59,13 @@ RUN apt-get -y update && \
 COPY build/configs/standalone/supervisor/ /etc/supervisor/conf.d/
 COPY --chmod=755 build/scripts/standalone/entrypoint.sh /entrypoint.sh
 
+# Give the 'ds' service user a writable HOME. supervisord runs as root and does
+# not reset HOME when dropping to user=ds, so without this the node services
+# inherit HOME=/root and fail to write their cache (e.g. sharp/pkg extracting
+# native modules to ~/.cache), disabling image processing. HOME is set per
+# program in the supervisor confs; this just ensures the directory exists.
+RUN mkdir -p /home/ds && chown ds:ds /home/ds
+
 #RUN mkdir -p ${EO_LOG}/docservice ${EO_LOG}/converter \
 #             ${EO_LOG}/adminpanel ${EO_LOG}/metrics
 
--- a/build/configs/standalone/supervisor/ds-adminpanel.conf
+++ b/build/configs/standalone/supervisor/ds-adminpanel.conf
@@ -2,7 +2,7 @@
 command=%(ENV_EO_ROOT)s/server/AdminPanel/server/adminpanel
 directory=%(ENV_EO_ROOT)s/server/AdminPanel
 user=ds
-environment=NODE_ENV=production-linux,NODE_CONFIG_DIR=%(ENV_EO_CONF)s,NODE_DISABLE_COLORS=1,APPLICATION_NAME=onlyoffice,LD_LIBRARY_PATH=%(ENV_EO_ROOT)s/server/FileConverter/bin:%(ENV_EO_ROOT)s/server/FileConverter/lib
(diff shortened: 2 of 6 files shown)


chrip (approved):
## Verification
| supervisord runs as root and doesn't reset HOME on `user=` drop | `supervisord.conf` has `user=root` and **no** global `environment=`; supervisord's documented behavior is that `user=` changes uid but does not set `HOME`. Programs therefore inherit `HOME=/root`. | ✓ |
| All `ds`-owned services get HOME | Exactly 5 confs use `user=ds` (adminpanel, converter, docservice, example, metrics) — the diff adds `HOME=/home/ds` to **all 5**. No ds-owned program left out. | ✓ |
| `/home/ds` directory created and writable by ds | Dockerfile L65: `RUN mkdir -p /home/ds && chown ds:ds /home/ds`. `mkdir` under root umask 022 → mode 755, then chowned to `ds:ds`. Writable by the service user, **not** world-writable. | ✓ |
| `chown ds:ds` resolves (ds user+group exist) | The `useradd` in the Dockerfile is commented out (L40); the `ds` user/group are created by the `.deb` install, which runs **before** the mkdir/chown step. CI build (amd64 + arm64, both flavors) succeeded → `chown` resolved. | ✓ |
| Consistent with standalone/orchestrated structure | `orchestrated.bake.Dockerfile` already uses `/home/ds` as the ds home (L37 `mkdir -p … /home/ds`, L44 `chown -R ds:ds /home/ds`, L332 `useradd … -d /home/ds …`). The standalone fix aligns with that existing convention. | ✓ |
| No conflict/duplication with entrypoint | `entrypoint.sh` does not set `HOME`, create `/home/ds`, or switch user; it `exec`s supervisord which performs the privilege drop. No overlap. | ✓ |
| No regression risk | Purely additive env var + one mkdir/chown; doesn't touch service commands, ports, or config. Adding HOME to ds-metrics (which doesn't use sharp) is harmless and keeps configs uniform. | ✓ |

## Issues & Suggestions

### 🔴 Blocking
None.

### ⚠️ Major
None.

### ℹ️ Minor / 💡 Suggestions
- ℹ️ **No automated regression guard for sharp loading.** The fix is verified manually (the PR's Test section: open a document with images, confirm the `Sharp module failed to load` / `processImageOptimal: Sharp module not available` warnings are gone). The e2e job is green but may not assert image optimization specifically, so a future silent regression (e.g. HOME dropped again) wouldn't be caught by CI. Acceptable for a Docker/supervisor env fix — flagging as a follow-up idea only, not a blocker.
- 💡 **Alternative: set HOME once, not five times.** The same effect could be achieved with a single global `environment=HOME=/home/ds` in `supervisord.conf` (or a shared include). The current per-program approach is arguably clearer and more targeted (it doesn't leak `HOME=/home/ds` onto any root-run program), so this is purely optional — no change requested.
- 💡 **Deb-created `ds` home is overridden.** The `.deb` likely gives `ds` its own home in `/etc/passwd` (not `/home/ds`); setting `HOME=/home/ds` explicitly overrides it. That's intentional and matches the orchestrated image, and nothing in the entrypoint relies on the original ds home — low risk, noted for completeness.


## Verdict
**Approve** — Correct root-cause diagnosis and a minimal, well-documented fix that gives the
`ds` services a writable HOME without introducing any world-writable directory; scope is
focused, it is consistent with the orchestrated image's existing `/home/ds` convention, and
all CI is green. The only notes are optional (an automated regression guard and an optional
single-point HOME definition).

Issue #78, opened 2026-04-18
Editor crashes with TypeError: Cannot read properties of undefined (reading '0') in updateHint


fredricadell:
## Environment                                                                                                                              
  - Euro-Office Document Server version: 9.2.1
  - Nextcloud version: latest Docker image                                                                                                    
  - ONLYOFFICE connector version: 9.2.1.37
  - Deployment: Docker (x86_64 native)        
                                                                                                                                              
  ## Description                              
  When opening any document (new or existing) via the Nextcloud ONLYOFFICE connector,                                                         
  the editor loads but immediately crashes with the following error:                 
                                                                                                                                              
  TypeError: Cannot read properties of undefined (reading '0')
    at n.updateHint (app.js:8:299088)                                                                                                         
    at n.onAppReady                                                                                                                           
                                                                                                                                              
  The document never becomes editable. The error appears in the Euro-Office docservice log at:                                                
  /var/log/onlyoffice/documentserver/docservice/out.log                                                                                       
                                                                                                                                              
  ## Additional info                          
  - Backend communication works correctly (all callbacks return {"error":0})                                                                  
  - The exact same setup with onlyoffice/documentserver works without any issues
  - The bug appears to be in the frontend JavaScript editor                                                                                   
                                                           
  ## Steps to reproduce                                                                                                                       
  1. Deploy Euro-Office Document Server via Docker                                                                                            
  2. Connect to Nextcloud via ONLYOFFICE connector                                                                                            
  3. Try to open any .docx file

rikled:
Thanks! Could you try with the latest nightly?

juliusknorr:
This should have been resolved. I'll close this for now, but feel free to reopen if this is still the case on the latest release

PR #232, opened 2026-06-23
feat: add front hot-reload development mode

Alex-Arsys:
Adds source-mode serving for web-apps and sdkjs so UI changes are visible with a browser reload (no grunt build). Includes auto-reload via a livereload server for web-apps (polling mode, reliable on WSL/Windows).

- develop/Makefile: add front-dev, front-prod, front-dev-live, sdkjs-dev targets with SDK sentinel guard, nginx validate-before-swap with rollback to .prod.bak, and automatic npm ci bootstrap on first run
- develop/setup/ds-docservice-dev.conf: nginx dev config serving web-apps and sdkjs from source (/develop), livereload injection via sub_filter, service worker neutralised, graceful fallback when livereload server is not running
- develop/setup/livereload-tools/: Node livereload server (polling 1s) with versioned lockfile for reproducible installs across machines

### DIFF
--- a/develop/Makefile
+++ b/develop/Makefile
@@ -19,6 +19,23 @@ endif
 HOST_LAN_IP ?=
 export HOST_LAN_IP
 
+# --- Front hot-reload (source mode) ---
+# All paths below are INSIDE the eo container. The repo root is bind-mounted at /develop
+# (docker-compose.yml: `../:/develop`), which is why source paths are /develop/develop/... .
+EO_REPO        := /develop
+LIVERELOAD_DIR := $(EO_REPO)/develop/setup/livereload-tools
+LIVERELOAD_SRV := $(LIVERELOAD_DIR)/server.js
+LIVERELOAD_PAT := $(LIVERELOAD_DIR)/[s]erver.js
+NGINX_CONF     := /etc/nginx/includes/ds-docservice.conf
+NGINX_CONF_BAK := $(NGINX_CONF).prod.bak
+DEV_CONF_SRC   := $(EO_REPO)/develop/setup/ds-docservice-dev.conf
+SDK_SENTINEL   := $(EO_REPO)/sdkjs/common/AllFonts.js
+# Canonical published docs (from documentation/brand.yml: brand.docs_url).
+DOCS_HOTRELOAD := https://docs.euro-office.com/development/hot-reload/
+
+.PHONY: local mobile next next-mobile pull wipe-nc wipe-next build \
+        front-dev front-prod front-dev-live sdkjs-dev refresh-urls print-nc-version
+
 local:
 	@echo ""
 	@echo "   ███████╗ ██████╗"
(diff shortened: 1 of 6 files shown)


MonaAghili:
@moodyjmz  — status check on `feat/front-hot-reload` against the two webpack migrations (web-apps, done; sdkjs, in flight).

## TL;DR

Works today, no action needed. One known, already-scoped break coming when sdkjs's webpack migration merges — a one-line Makefile fix, not a redesign.

## web-apps: still fine, verified

`develop/setup/ds-docservice-dev.conf` serves `/web-apps/` straight from the source tree (`alias /develop/web-apps/`). That depends on the *un-built* `index.html` loading `require.js` → `app_dev.js` — a plain RequireJS `require.config`/`require([...])` listing every individual source module.

web-apps' Grunt→webpack migration (`74a13d1515`, already merged) only touches the production/deploy path: it builds `index.html.deploy` into a bundled `app.js` under `BUILD_ROOT`. The dev entry point (`index.html`/`app_dev.js`) and the production entry point (`index.html.deploy`) are separate files — webpack landing never touched the one hot-reload depends on. Nothing to change here.

## sdkjs: fine today, breaks on merge — fix already known

`develop/Makefile:199` (`sdkjs-dev` target) calls `grunt develop` to generate the SDK's unbundled dev loader (`sdkjs/develop/sdkjs/*/scripts.js` + `AllFonts.js`). That's correct as long as sdkjs main still has Grunt — but `feature/migration-to-webpack-sdkjs` removes it, so once that branch merges, `sdkjs-dev` fails outright (`grunt: command not found`).

I checked what that branch actually ships before flagging this as a real problem, and it isn't one — it's a rename, not a redesign:

- `build/package.json` on that branch: `"develop": "node scripts/build-develop.cjs"`
- Same output path: `develop/sdkjs/{word,cell,slide,visio}/scripts.js`
- Its own `build/DEVELOPER-GUIDE.md` documents pairing `npm run develop` with our `make front-dev` / `front-dev-live` / `front-prod` directly — the migration was written with this hot-reload workflow in mind, not against it.

So the fix, to land in the same PR that merges sdkjs's webpack migration:

```diff
  # develop/Makefile:199
- cd $(EO_REPO)/sdkjs/build && npm install --no-audit --no-fund && grunt develop
+ cd $(EO_REPO)/sdkjs/build && npm install --no-audit --no-fund && npm run develop
```

No other changes needed — `SDK_SENTINEL` and the `[s]erver.js` livereload-process check both key off paths that don't move.

## Ask

Can you (or whoever merges `feature/migration-to-webpack-sdkjs`) carry that one-line Makefile change along with it? I can open it as a follow-up PR once that branch is closer to merging, or fold it into the same PR — whichever's easier for you.

j-base64:
I'm taking a look 👀

j-base64:
First pass review @MonaAghili,

### Summary 
This PR introduces the beginning of a solution for an interesting problem (more a livereload experience than a hot reload one), but looks incomplete and mostly tested with web-apps only (very partial since sdkjs is a big chunk of the daily dev scenario).

This work probably needs deeper testing and finetuning by frontenders themselves, along at least a monthly, real-world dev journey, to catch issues that will surface over the long run.

Also, the author looks inactive for now, and we've set up similar tooling since this PR landed. Unless someone else is interested in improving this, to avoid too many open PRs by us that we can't actually treat as priority,
**maybe we can close it as-is for now and see later if it's worth reopening to improve or go on a different path (see full review below).**

> NOTE: This PR was created for a goal we identified internally,
> but no issue was opened to document what triggered it. Also, since this PR is more a frontend tool than a server one, the review route chosen
> probably made it land here silently.


<details>
<summary><strong>📜 VIEW FULL REVIEW (+ possible alternatives and benchmark)</strong></summary>

## Intro
The goal this PR tries to achieve is interesting: **get closer to an HMR-like dev experience** and reduce time from frontend codebase edit to browser render.

"front hot reload" is a bit oversold since it's livereload here (not HMR), but at the same time, the current architecture doesn't support module-level hot-swap.

Its reliance on serving source was benchmarked against serving freshly rebuilt deploy files on file changes - see the dev-loop latency benchmark below.

## Observations on this PR (partial, need more testing):

The mechanism under the hood looks like: livereload injection on server + serving source files + file-change detection that retriggers livereload.

The code itself doesn't show big issues at a glance (this review didn't focus a lot on the code itself).

The work seems to have been mostly focused on web-apps, with sdkjs support feeling more like an option bolted on afterward:

- 🟠**Sdkjs, that is an important part of the project for frontenders, is disabled in the code**, with a comment saying it's unreliable across the WSL/Windows filesystem boundary.
  - tried uncommenting the one line that enables it and restarting - it worked through a couple of fast tests (a sdkjs edit triggered a browser reload), but wasn't deeply tested. Only tried in this one setup though (VS Code Remote-WSL, so the file write happens natively inside WSL) - not proof this works for every contributor's editing setup, just that it's not broken for this one.
- separately, this PR's `sdkjs-dev` target still calls `grunt develop`, which depends on sdkjs's build/Gruntfile.js. @MonaAghili's earlier comment already flagged this and proposed a one-line fix (`npm run develop`).

🟠 **First quick tests pinpointed an issue that could be the beginning of other findings** that this PR should cover:
  - less.js is retriggered client-side at each page load whether or not you changed a css file, adding a bit to every reload regardless (folded into the Source-serving row of the benchmark below).
  - not verified yet how this PR deals with isolated html changes or assets like svg or binaries (pictures etc...), service workers updates...

## Alternative or complementary approach to consider

- relying on rebuilding isolated deploys on file change instead of serving source files (connecting to what webpack watch now offers)
- living outside documentserver's repo instead of inside it (scripts overriding the server or docker image once it's up)

Experimented with this build-on-file-changes approach independently of this PR - works almost well, but build times still need improvement (details in the benchmark below).

### 🧪 Latency benchmark: 2 methods measured

| Approach | ETA (5-trial avg, steady-state) | Pro | Con |
|---|---|---|---|
| **Source-serving** (this PR's own tooling) | ~5.7s | Quick results in the browser | Performance not evaluated in a scope close to the end-user one; doesn't limit reprocessing to only what actually changed (eg. less.js recompiles every time even for only JS edits) |
| **Build** (webpack watch + browsersync) | ~11.7s | Closer to what we ship (better for tests); can target a specific editor/platform | sdkjs build time still long but potentially improvable (see issue [#78](https://github.com/Euro-Office/sdkjs/issues/78), which also hopes for code splitting down the line) |

#### Test protocol used (same for both approaches):
Same single-line, highly visible sdkjs edit for both (header-corner fill color override, forced to solid red - renders on any blank spreadsheet, no content needed), timed on one already-open browser tab per method from file save to that tab finishing its own auto-refresh (real reload, real render - not a fresh page load per edit), 5 trials per method, averaged as a steady-state figure excluding a cold-start first trial.

## To go forward

This is a devtool that needs to evolve during the whole life of the project (support service workers etc...), preferably driven by frontend developers using it and with a great overview on how this works.

One direction worth considering: make it a config toggle allowing developers to switch between live-source mode (fast, this PR's approach) and live-rebuild mode (closer to what actually ships) depending on what they're working on at that moment, instead of committing the whole project to one approach.

## Verdict:
I'm interested in using or contributing to this kind of tool, and can help with similar tooling already done on the side (built method, scripts aside overriding docker image). Unless someone else is interested in improving this now - and to avoid open PRs by us that we can't treat as priority - **I suggest closing it as-is for now and see later if it's worth reopening to improve or go on a different path.**


</details>

Issue #13, opened 2026-03-28
How to rebuild the image wthout SSE4.2 CPU instruction support?


oleua:
Hi! I have quite old CPU which doesn't support SSE4.2 CPU instruction, which is demanded by sharp  https://github.com/lovell/sharp/issues/3743

So, your image is not working for me sadly - this is what I am getting in the /var/log/onlyoffice/documentserver/docservice/out.log:
``` 
[2026-03-28T10:22:58.910] [WARN] [localhost] [docId] [userId] nodeJS - Sharp module failed to load. Image processing functionality will be limited.
[2026-03-28T10:22:58.915] [WARN] [localhost] [docId] [userId] nodeJS - Sharp load error: 
Something went wrong installing the "sharp" module

EACCES: permission denied, mkdir '/root/.cache/pkg/282c93e4113a05841f827943c9f6cd6d4188bfda819ab562b27b278d2fcebfda'

Possible solutions:
- Install with verbose logging and look for errors: "npm install --ignore-scripts=false --foreground-scripts --verbose sharp"
- Install for the current linux-x64 runtime: "npm install --platform=linux --arch=x64 sharp"
- Consult the installation documentation: https://sharp.pixelplumbing.com/install
```
This influence on Fonts Generation, creating previews etc, so the document server is not working. 

Is it possible to overcome this limitation somehow in your image?

oleua:
I've checked again and found that in the recent docker image versions (latest and latest-dev) this trouble persists:
```
Starting program: /var/www/onlyoffice/documentserver/server/DocService/docservice 
warning: Error disabling address space randomization: Operation not permitted
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7187966006c0 (LWP 2945)]
[New Thread 0x718795c006c0 (LWP 2946)]
[New Thread 0x7187952006c0 (LWP 2947)]
[New Thread 0x71878fe006c0 (LWP 2948)]
[New Thread 0x71878f4006c0 (LWP 2949)]
[New Thread 0x71876e8006c0 (LWP 2950)]
[New Thread 0x71876de006c0 (LWP 2951)]
[New Thread 0x71876d4006c0 (LWP 2952)]
[New Thread 0x71876ca006c0 (LWP 2953)]

Thread 1 "docservice" received signal SIGILL, Illegal instruction.
0x000071875b130d83 in ?? ()
[...]

PR #238, opened 2026-06-24
chore: replace short JWT_SECRET defaults with ≥32-char value

chrip:
## Context

Companion PR to Euro-Office/document-server-integration#13. **Both PRs must be merged together** — this PR bumps the `document-server-integration` submodule pointer to the commit in that PR.

The EuroOffice connector v11+ enforces a minimum JWT secret length of 32 characters. Short defaults in example commands and the orchestrated entrypoint fallback cause an immediate runtime error: *"JWT secret key is too short (minimum 32 characters required)"*.

## What changes

- `README.md` — docker run example secret updated to ≥32 chars
- `build/scripts/orchestrated/docker-entrypoint.sh` — `${JWT_SECRET:=secret}` fallback raised to 35 chars
- `build/scripts/orchestrated/example-docker-entrypoint.sh` — same
- `document-server-integration` — submodule bumped to Euro-Office/document-server-integration#13

Assisted-by: ClaudeCode:claude-sonnet-4-6

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ We currently provide a docker image for testing and integration purposes. We are
 
 ```
 docker pull ghcr.io/euro-office/documentserver:latest
-docker run -i -t -d -p 8080:80 --restart=always -e EXAMPLE_ENABLED=true -e JWT_SECRET=my_jwt_secret ghcr.io/euro-office/documentserver:latest
+docker run -i -t -d -p 8080:80 --restart=always -e EXAMPLE_ENABLED=true -e JWT_SECRET=my_long_jwt_secret_at_least_32_chars ghcr.io/euro-office/documentserver:latest
 ```
 
 ### Building
--- a/build/scripts/orchestrated/docker-entrypoint.sh
+++ b/build/scripts/orchestrated/docker-entrypoint.sh
@@ -135,7 +135,7 @@ export NODE_CONFIG='{
       },
       "secret": {
         "inbox": {
-          "string": "'${JWT_SECRET_INBOX:-${JWT_SECRET:=secret}}'"
+          "string": "'${JWT_SECRET_INBOX:-${JWT_SECRET:=euro-office-dev-jwt-secret-key-2026}}'"
         },
         "outbox": {
           "string": "'${JWT_SECRET_OUTBOX:-${JWT_SECRET}}'"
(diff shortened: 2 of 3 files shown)


MonaAghili (commented):
Verified against both PR branches directly (not just the diffs).

**Correctness:** the touched paths do fix the crash — both new literals are ≥32 chars (`my_long_jwt_secret_at_least_32_chars` = 36, `euro-office-dev-jwt-secret-key-2026` = 35).

**Two things worth resolving before merge:**

1. **#13's sweep is incomplete.** `git diff main pr13 --name-only` shows only 9 files changed, but these still have secrets under 32 chars and will still reproduce the original error:
   - `web/documentserver-example/nodejs/config/default.json:37` — `"secret": "secret"`. This is what node-config actually loads when `NODE_CONFIG` isn't set (e.g. local `npm start` outside Docker).
   - `web/documentserver-example/csharp/README.md:47`, `csharp-mvc/README.md:46` — `value="secret"`.
   - `nodejs/README.md:55,176` inline JSON examples, plus the documented-default tables in `php`, `php-laravel`, `python`, `ruby` READMEs — all still show `secret` / `your-256-bit-secret`.

2. **This trades a loud failure for a silent one.** DocumentServer's production `docker-entrypoint.sh` enables JWT by default (`JWT_ENABLED:=true`). Previously, forgetting to set `JWT_SECRET` crashed immediately — an accidental safety net. Now the fallback is long enough to pass validation, so that same misconfiguration starts up fine, silently signing with a secret that's public in this repo's git history. Worth either failing startup when `JWT_ENABLED=true` and `JWT_SECRET` is unset, or at least logging a loud warning when the fallback is used.

No other issues — quoting/expansion in the entrypoint scripts is unchanged, and the submodule bump points at exactly #13's head commit with nothing extra riding along.

MonaAghili (approved):
One residual, low-severity note (not blocking): the standalone `nodejs` example's own `docker-entrypoint.sh` still falls back to the new long literal when `JWT_SECRET` is unset — but `JWT_ENABLED` there defaults to `false`, so it can't silently turn on JWT the way the DocumentServer production entrypoint could. This is a test/example client, not the server, so I wouldn't hold up the merge for it.

No other issues found — quoting/expansion is unchanged, and DCO/clone checks pass. The only non-green check is a **cancelled** (not failed) build run against the current head commit; it hasn't actually re-run since the last push, so I'd re-trigger CI before merging rather than merge on stale results.

PR #249, opened 2026-06-26
use service to start rabbitmq-server

BlackIkeEagle:
this makes it consistent with postgresql and redis-server


Alex-Arsys:
Locally tested. LGTM. 

Maybe I only can say commit message doesn't follow conventional commits but the rest is ok.

BlackIkeEagle:
> Locally tested. LGTM.
> 
> Maybe I only can say commit message doesn't follow conventional commits but the rest is ok.

ok added `chore:` at the start

chrip (approved):
LGTM :)

Issue #245, opened 2026-06-25
Dev env: `make local` hangs silently when Nextcloud install stalls (no timeout, no diagnostics)


moodyjmz:
**TL;DR:** `make local` (dev env) can hang for **hours** behind a single uninformative `[nextcloud] still installing (Ns)` line — a ~20.6h hang (74246s) was just reported. The `refresh-urls` wait loop has **no timeout** (it only bails if the container *stops*) and conflates every failure mode into one message, so the actual cause is invisible. Propose: a progress-aware timeout with actionable output, phase-aware status, proactive stale-volume detection, and a `make doctor` diagnostic — factored into a shared helper so the incoming `eo.sh` (#186) benefits too.

<details>
<summary>The problem</summary>

`make local` → `docker compose up -d` → `refresh-urls`. The Nextcloud wait loop (`develop/Makefile:131-145`) polls:
```sh
until occ config:app:get eurooffice DocumentServerUrl | grep -q "http"; do … "still installing (Ns) — <last log>" … done
```
It only exits when the `eurooffice` app is enabled + `DocumentServerUrl` is set (written by `develop/setup/shared/hooks/after-install.sh`). With **no timeout**, any stalled install loops forever. Compounding: the `nextcloud` service has **no healthcheck** (`develop/docker-compose.yml:25`) so readiness is inferred from `State.Status == running` (true long before install finishes); and `after-install.sh` blocks on `until curl http://eo/healthcheck` *before* `occ app:enable`, under `set -e` (a failing `occ` step aborts the hook silently).
</details>

<details>
<summary>Failure modes the one message masks (today they look identical)</summary>

1. **Stale / half-installed data volume** — the official image runs `after-install` hooks only on a *fresh* install; a prior incomplete run leaves a volume that skips the hook → app never enabled → infinite loop. **Most common.**
2. **`after-install.sh` blocked on eo** — the nextcloud container can't reach `http://eo/` (cross-container DNS/network), though the Makefile's eo gate passed (it curls from *inside* eo). Telltale `still waiting...` buried in NC logs.
3. **`occ app:enable eurooffice` fails** (autoloader/app-dir) → `set -e` aborts the hook with no surfaced error.
4. **Image behind the data volume** — the existing `local` check only runs when the container is *not* running (`Makefile:~43`), so it slips through when up.
</details>

<details>
<summary>Proposed improvements (prioritized)</summary>

**MVP**
[...]

PR #252, opened 2026-06-26
fix: Adding SELinux flags for compose file volumes

Aiiaiiio:
`enable-office.sh` wasn't able to run during setup, because the executable flags won't get carried over by default if SELinux is present. This add the :z flag for all mounts.

Fixes a part of: https://github.com/Euro-Office/DocumentServer/issues/245

### DIFF
--- a/develop/docker-compose.yml
+++ b/develop/docker-compose.yml
@@ -20,8 +20,8 @@ services:
       JWT_SECRET: ${EO_JWT_SECRET:-euro-office-dev-jwt-secret-key-2026}
       THEME: euro-office
     volumes:
-      - ../:/develop # mounts deployed files to application
-      - ./setup/Makefile:/Makefile # mounts deployed files to application
+      - ../:/develop:z # mounts deployed files to application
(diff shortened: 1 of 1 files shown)


moodyjmz (approved):
Verified on macOS Docker Desktop: `docker info` reports no SELinux support, so `:z` is a no-op on non-SELinux hosts (macOS, Windows WSL2, Ubuntu/AppArmor). No regression risk for developers not on Fedora/RHEL. Fix is correct and coverage is complete — all four bind mounts updated, `nc_data` named volume correctly left alone.

Two minor notes for the record:

- The PR description says "executable flags won't get carried over" — the execute bit is actually preserved by bind mounts. What SELinux blocks is access based on the file's security context label (`user_home_t` etc.), which `:z` fixes by relabeling to `container_share_t`. Doesn't affect correctness, but worth knowing for future debugging.
- `Fixes: #245` overstates scope a little — issue #245 covers several failure modes (stale volume, timeout, diagnostics); this PR addresses the specific one where SELinux denies execution of the post-install hook. The issue will stay open for the rest.

PR #274, opened 2026-07-03
Fix Application meta tag in saved documents

### LINKED ISSUE #273: Wrong app name in the app field pf a document metatag
Hi! The meta tags contain wrong information about the app the document was created. Should be euro-office.

Image

Mumrah213:
Fixes #273

### DIFF
--- a/build/configs/standalone/supervisor/ds-adminpanel.conf
+++ b/build/configs/standalone/supervisor/ds-adminpanel.conf
@@ -2,7 +2,7 @@
 command=%(ENV_EO_ROOT)s/server/AdminPanel/server/adminpanel
 directory=%(ENV_EO_ROOT)s/server/AdminPanel
 user=ds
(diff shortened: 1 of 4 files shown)


chrip (approved):
Thanks @Mumrah213 — clean, well-scoped fix, and I'm happy to approve it. 👍

It does the thing that matters: the standalone converter now writes **`euro-office`** into the document's `Application` metadata instead of the old `onlyoffice`, matching what the orchestrated Docker build already produces. I verified x2t writes this value verbatim into `docProps/app.xml` (and the equivalent field for the other formats), so the fix is correct.

For the record, this surfaces a broader inconsistency — the same field is populated from different sources across deployments and they don't all agree yet:
- Docker (standalone + orchestrated): `euro-office`
- systemd / RPM packages: `EURO-OFFICE`
- Windows installer: still **`ONLYOFFICE`** (`sCompanyName` in `defines.iss`)

That's **not a blocker for this PR** — I'll open a separate issue to align them. 

Thanks again for cleaning this up!

PR #301, opened 2026-07-27
feat(standalone): configurable file-size limits + Postgres schema bootstrap

ckbaker10:
## Problem

Two unrelated gaps in `build/scripts/standalone/entrypoint.sh`, bundled here because they touch
the same file:

1. Upload/conversion size limits (`limits_tempfile_upload`, `maxDownloadBytes`, the two
   `inputLimits[].zip.uncompressed` values, nginx's `client_max_body_size`) are hardcoded. Raising
   them today means `sed`-patching files inside a running container, which isn't under any
   documented volume — the edit is gone on next `docker run`/recreate.
2. A fresh Postgres volume leaves docservice failing with `DB table "task_result" does not exist`
   and never binding `:8000`, so nginx serves 502 on `/healthcheck` until someone applies the
   schema by hand.

## Approach

**Size limits** — four new env vars, defaults match current shipped values (unset = no behavior
change):

| Var | Default | Applies to |
|---|---|---|
| `MAX_FILE_SIZE` | `104857600` | `services.CoAuthoring.server.limits_tempfile_upload`, `FileConverter.converter.maxDownloadBytes` (`local.json`), example app's `maxFileSize` |
| `MAX_ZIP_UNCOMPRESSED_DOC` | `50MB` | `inputLimits[].zip.uncompressed` for docx/pptx/vsdx |
| `MAX_ZIP_UNCOMPRESSED_XLSX` | `300MB` | same, xlsx |
| `NGINX_CLIENT_MAX_BODY_SIZE` | `100m` | `client_max_body_size` in `ds-common.conf` |

`inputLimits` gets patched with `sed` directly on `default.json` rather than merged into
`local.json` — node-config replaces arrays wholesale instead of merging by element, which would
silently drop the `template` field from every entry. `ds-example.conf` needs no change, it has no
`client_max_body_size` of its own and inherits from `ds-common.conf`'s scope.

**Schema bootstrap** — `ensure_db_schema()`, called after `service postgresql start`: waits for
Postgres, checks for the `task_result` table, applies `server/schema/postgresql/createdb.sql` if
missing. No-op on an already-initialized DB.

## Testing

- jq filter logic checked against a fixture matching `server/Common/config/default.json`'s actual
  schema: numeric coercion and per-family zip-limit targeting both correct.
- Verified on two live instances:
  - `client_max_body_size` in `ds-common.conf` actually gets patched and nginx accepts the larger
    body size.
  - Uploading a file larger than the old 100MB default through `/example/` succeeds end-to-end.
  - Fresh Postgres volume: `ensure_db_schema` creates the schema on first boot, docservice comes
    up healthy instead of 502-ing on `/healthcheck`.
  - New env vars left unset: config/behavior matches what shipped before this change.

## AI disclosure

Drafted with AI assistance (Claude Code / Sonnet 5) per the repo's AI Contribution Policy.

### DIFF
--- a/build/scripts/standalone/entrypoint.sh
+++ b/build/scripts/standalone/entrypoint.sh
@@ -16,6 +16,7 @@ EXAMPLE_LOCAL="${EXAMPLE_CONF_DIR}/local.json"
 NGINX_CONFIG_PATH="/etc/nginx/nginx.conf"
 NGINX_DS_DIR="${EO_CONF}/nginx"
 NGINX_DS_CONF="${NGINX_DS_DIR}/ds.conf"
+NGINX_DS_COMMON_CONF="${NGINX_DS_DIR}/includes/ds-common.conf"
 NGINX_DS_SSL_TMPL="${NGINX_DS_DIR}/ds-ssl.conf.tmpl"
 SUPERVISOR_CONF_DIR="/etc/supervisor/conf.d"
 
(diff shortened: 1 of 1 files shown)


chrip (changes requested):
Thanks @ckbaker10! Requesting changes — nothing wrong with the code itself: the schema-bootstrap half is genuinely good and approvable on its own, and I found no correctness bug in the paths I could verify statically. Two things before merge, though:

- **It collides with #237**, which is still open and touches the same `build/scripts/standalone/entrypoint.sh` for the same settings (`.FileConverter.converter.maxDownloadBytes` and the zip `inputLimits`) under different env-var names and a different mechanism. Merging both would conflict and leave two competing controls for one setting.
- **It bundles an unrelated concern** — the size limits and the Postgres schema bootstrap are two separate changes sharing a file.

Could you **split out `ensure_db_schema`** into its own PR (I'd merge that quickly), and **reconcile the size-limit half with #237** — one env-var scheme and one mechanism per setting? Happy to help align the naming.


ckbaker10:
closed in favour of

https://github.com/Euro-Office/DocumentServer/pull/304
https://github.com/Euro-Office/DocumentServer/pull/305

PR #332, opened 2026-08-18
fix(docker): start rabbitmq directly and wait for readiness

### LINKED ISSUE #326: Docker latest stuck on RabbitMQ startup, direct rabbitmq-server works
Hi,

I think there is an issue with the current `ghcr.io/euro-office/documentserver:latest` image.

I’m running it with Docker Compose for Nextcloud integration.

The container starts PostgreSQL correctly, then gets stuck here:

```text
Starting PostgreSQL 16 database server
...done.
Postgres schema already present, skipping.
Starting RabbitMQ Messaging Server rabbitmq-server
```

After that, nothing else starts.

The healthcheck fails with:

```text
curl: (56) Recv failure: Connection reset by peer
```

I checked the container and `start-stop-daemon` stays at almost 100% CPU:

```text
start-stop-daemon --quiet --chuid rabbitmq \
  --start --exec /usr/sbin/rabbitmq-server \
  --pidfile /var/run/rabbitmq/pid --background
```

At this point RabbitMQ is not really started:

```text
rabbitmqctl status
```

fails, and `epmd -names` only shows the rabbitmq CLI node.

No listener on `5672`.

The interesting part is that RabbitMQ itself works fine if I start it manually inside the container:

```bash
su rabbitmq -s /bin/sh -c "/usr/lib/rabbitmq/bin/rabbitmq-server"
```

or:

```bash
su rabbitmq -s /bin/sh -c \
"RABBITMQ_PID_FILE=/var/run/rabbitmq/pid /usr/lib/rabbitmq/bin/rabbitmq-server -detached"
```

Then:

```text
rabbitmqctl status
```

works correctly and ports `5672` and `25672` are listening.

I checked `/entrypoint.sh` and it currently starts RabbitMQ with:

```bash
service rabbitmq-server start
```

This call seems to be the part that hangs.

As a workaround, I built a small custom image where I replaced:

```bash
service rabbitmq-server start
```

with a direct detached RabbitMQ startup, then wait for:

```bash
rabbitmqctl status
```

before continuing.

With this workaround the container starts normally:

```text
Starting RabbitMQ directly
...done.
Starting redis-server: redis-server.
Starting nginx nginx
...done.
```

and later:

```text
converter entered RUNNING state
docservice entered RUNNING state
```

The healthcheck also works again:

```text
HTTP/1.1 200 OK

true
```

Image used:

```text
ghcr.io/euro-office/documentserver:latest
```

Created:

```text
2026-08-04T08:22:40.568547219Z
```

Digest:

```text
sha256:6068075cc247f9283d337b2cf9a105d9df49fcd8cd6b5e82fa90640df1552e80
```

I also noticed RabbitMQ data already present in the image under:

```text
/var/lib/rabbitmq/mnesia/rabbit@buildkitsandbox
```

I don’t know if it is related, but I thought it was worth mentioning.

Let me know if you want more logs or tests.

chrip:
PR #249 switched the standalone entrypoint from a direct detached start to `service rabbitmq-server start` for consistency with postgres and redis. The init script goes through `start-stop-daemon --background`, which wedges under a very large RLIMIT_NOFILE. Docker daemons that pass one through to containers (26.1.5+dfsg1 defaulting to 1073741816 was reported twice) leave the entrypoint stuck before nginx and supervisord ever start: the container looks hung, /healthcheck resets the connection, and nothing in the log says why. Setting `ulimits.nofile` in compose is a workaround, but users have no way to know that is what is wrong.

Go back to `rabbitmq-server -detached`, which is unaffected, and follow it with a bounded `rabbitmqctl await_startup` poll (AMQP_START_TIMEOUT, default 60s). `-detached` returns before the node registers with epmd, so a single await_startup call fails outright and has to be polled. A timeout now logs rabbitmqctl's own diagnostics and continues to supervisord, like the schema bootstrap above it, so the failure is visible instead of silent.

Apply the same at image build time, where the identical hang would break builds on affected hosts, and drop the mnesia database the build leaves behind: the node name follows the container hostname, so the baked `rabbit@buildkitsandbox` directory is never read at runtime.

Fixes #326

### DIFF
--- a/build/.docker/standalone.bake.Dockerfile
+++ b/build/.docker/standalone.bake.Dockerfile
@@ -40,12 +40,19 @@ RUN apt-get update && \
 #RUN useradd -r -s /bin/false ds || true
 
 # --- install ${COMPANY_NAME_LOW} .deb package
+# RabbitMQ is started directly rather than via `service rabbitmq-server start`
+# for the same reason as in entrypoint.sh: the init script's
+# `start-stop-daemon --background` wedges under a huge RLIMIT_NOFILE, which
+# would hang the build on hosts that hand containers one (#326). Its mnesia
+# database is dropped again afterwards -- the node name follows the container
+# hostname, so `rabbit@buildkitsandbox` is never read at runtime.
 ARG TARGETARCH
 COPY --from=packages / /tmp/
 RUN apt-get update && \
     (pg_createcluster 16 main || true) && \
     service postgresql start && \
-    service rabbitmq-server start && \
+    runuser -u rabbitmq -- rabbitmq-server -detached && \
+    runuser -u rabbitmq -- rabbitmqctl await_startup --timeout 60 && \
     sudo -u postgres psql -c "CREATE USER eurooffice WITH password 'eurooffice';" && \
     sudo -u postgres psql -c "CREATE DATABASE eurooffice OWNER eurooffice;" && \
     echo "${COMPANY_NAME_LOW}-${PRODUCT_NAME_LOW} ds/db-type string postgres" | debconf-set-selections && \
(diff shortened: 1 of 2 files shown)


rikled:
Not sure, but the failed builds appear to be related to the changes.

chrip:
The wait during the image build asked RabbitMQ "are you ready?" (`rabbitmqctl await_startup`) exactly once.
RabbitMQ starts in the background and needs a moment to announce itself before
anything can reach it, so that single question came too early and failed instantly
instead of waiting — which broke three of the four builds. The container startup
script already asks in a loop for this reason; the image build now does the same,
with a time limit so a genuinely broken RabbitMQ still fails the build and prints
why.

rikled (approved):
lgtm

AMQP_START_TIMEOUT is unvalidated. Since it's potential user input maybe we should check the value more, like we did with FILECONVERTER_MAX_DOWNLOAD_BYTES.

PR #350, opened 2026-08-28
ci: gate on sdk-all.js compiling under the real no-ICU V8

moodyjmz:
## Summary

- Adds a `sdkjs-v8-smoke` job that runs the reproduction recipe from
  Euro-Office/sdkjs#80 as a CI gate: pull the built image, run
  `x2t -create-js-cache` for word/cell/slide/visio, fail if any resulting
  `sdk-all.cache` is empty.
- Extracts the tag-resolution + GHCR login + pull sequence (previously only
  in `e2e`) into `.github/actions/pull-built-image`, since this PR needed
  the same sequence a second time and copy-pasting it again seemed like the
  wrong move.

## Why

#80: a Terser change emitted bare astral-plane object keys in `sdk-all.js`.
Node/Chromium (both full-ICU) parse them fine, but `x2t`/`doctrenderer`
embed a V8 built with `v8_enable_i18n_support=false`, which rejects them —
every document save aborted. `check-build.yml` validates bundles with
Node and headless Chromium only, so nothing in CI could have caught this.
This turns the incident's own manual repro into an automated gate, so a
future regression of the same class fails loud in seconds with a pointer
back to #80, instead of surfacing later as an unexplained save failure.

## Test plan

- [x] Both YAML files validated (`yaml.safe_load`)
- [ ] CI run on this PR — `sdkjs-v8-smoke` should pass against the current
      good bundle; will report back once it's run
- [ ] Reviewer: worth double-checking the `LD_LIBRARY_PATH`/path assumptions
      (`/var/www/euro-office/documentserver/...`) still match the current
      image layout

Assisted-by: ClaudeCode:claude-sonnet-5

### DIFF
--- a/.github/actions/pull-built-image/action.yml
+++ b/.github/actions/pull-built-image/action.yml
@@ -1,0 +1,49 @@
+name: 'Resolve and pull built image'
+description: >
+  Resolves the DocumentServer image tag for the current event (PR cache tag
+  on pull_request, :latest on a release tag, :nightly otherwise), logs in to
+  GHCR, and pulls it. Shared by any job that runs something against the
+  already-built image (e2e, post-build smoke checks, ...) so the
+  tag-resolution + login + pull sequence exists in exactly one place.
+
+inputs:
+  registry:
+    description: 'Container registry host, e.g. ghcr.io/euro-office'
+    required: true
+  image-name:
+    description: 'Image name used for :nightly / :latest tags'
+    required: true
+  cache-image-name:
+    description: 'Image name used for the per-PR cache tag'
+    required: true
+
+outputs:
+  ref:
+    description: 'Fully-qualified image ref that was pulled'
+    value: ${{ steps.image.outputs.ref }}
+
+runs:
+  using: composite
+  steps:
+    - name: Determine image
+      id: image
+      shell: bash
+      run: |
+        if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+          echo "ref=${{ inputs.registry }}/${{ inputs.cache-image-name }}:pr-${{ github.event.pull_request.number }}" >> "$GITHUB_OUTPUT"
+        elif [[ "${{ github.ref }}" == refs/tags/* ]]; then
+          echo "ref=${{ inputs.registry }}/${{ inputs.image-name }}:latest" >> "$GITHUB_OUTPUT"
+        else
+          echo "ref=${{ inputs.registry }}/${{ inputs.image-name }}:nightly" >> "$GITHUB_OUTPUT"
+        fi
+
+    - name: Log in to GitHub Container Registry
+      uses: docker/login-action@c99871dec2022cc055c062a10cc1a1310835ceb4 # v4.3.0
+      with:
+        registry: ${{ inputs.registry }}
+        username: ${{ github.actor }}
+        password: ${{ github.token }}
+
+    - name: Pull built image
+      shell: bash
+      run: docker pull "${{ steps.image.outputs.ref }}"
--- a/.github/actions/pull-built-image/action.yml
+++ b/.github/actions/pull-built-image/action.yml
@@ -46,4 +46,6 @@ runs:
 
     - name: Pull built image
       shell: bash
-      run: docker pull "${{ steps.image.outputs.ref }}"
+      env:
+        IMAGE_REF: ${{ steps.image.outputs.ref }}
+      run: docker pull "$IMAGE_REF"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -461,26 +461,13 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: Determine image
-        id: image
-        run: |
-          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
-            echo "ref=${{ env.REGISTRY }}/${{ env.CACHE_IMAGE_NAME }}:pr-${{ github.event.pull_request.number }}" >> $GITHUB_OUTPUT
-          elif [[ "${{ github.ref }}" == refs/tags/* ]]; then
-            echo "ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:latest" >> $GITHUB_OUTPUT
-          else
-            echo "ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:nightly" >> $GITHUB_OUTPUT
-          fi
-
-      - name: Log in to GitHub Container Registry
-        uses: docker/login-action@c99871dec2022cc055c062a10cc1a1310835ceb4 # v4.3.0
+      - name: Pull built image
+        id: pull
+        uses: ./.github/actions/pull-built-image
         with:
           registry: ${{ env.REGISTRY }}
-          username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Pull built image
-        run: docker pull ${{ steps.image.outputs.ref }}
+          image-name: ${{ env.IMAGE_NAME }}
+          cache-image-name: ${{ env.CACHE_IMAGE_NAME }}
 
       - name: Set up Node.js
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
@@ -500,7 +487,7 @@ jobs:
       - name: Run E2E tests
         working-directory: e2e
         env:
-          E2E_IMAGE: ${{ steps.image.outputs.ref }}
+          E2E_IMAGE: ${{ steps.pull.outputs.ref }}
         run: npm test
 
       - name: Upload Playwright report
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -497,3 +497,61 @@ jobs:
           name: playwright-report-${{ github.sha }}
           path: e2e/playwright-report/
           retention-days: 14
+
+  # Asserts the shipped sdk-all.js bundles actually compile under the real,
+  # no-ICU V8 that x2t/doctrenderer embeds — not just under Node/Chromium
+  # (both full-ICU, so CI JS checks structurally cannot catch this class of
+  # bug). This is the reproduction recipe from Euro-Office/sdkjs#80 (Terser
+  # emitted bare astral-plane object keys; the no-ICU V8 rejected them and
+  # x2t aborted on every save) turned into a CI assertion. Runs against the
+  # same already-built image as e2e, independently of it, so a V8-compile
+  # regression fails loud here in seconds rather than surfacing as a broken
+  # save with no diagnostic.
+  sdkjs-v8-smoke:
+    runs-on: self-hosted
+    needs: build
+    if: |
+      !cancelled() &&
+      needs.build.result == 'success' &&
+      (github.event_name != 'pull_request' ||
+       github.event.pull_request.head.repo.full_name == github.repository)
+    permissions:
+      contents: read
+      packages: read
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Pull built image
+        id: pull
+        uses: ./.github/actions/pull-built-image
+        with:
+          registry: ${{ env.REGISTRY }}
+          image-name: ${{ env.IMAGE_NAME }}
+          cache-image-name: ${{ env.CACHE_IMAGE_NAME }}
+
+      - name: Compile sdk-all.js under the real no-ICU V8 (x2t -create-js-cache)
+        run: |
+          docker run --rm --entrypoint bash "${{ steps.pull.outputs.ref }}" -c '
+            set -euo pipefail
+            DIR=/var/www/euro-office/documentserver
+            export LD_LIBRARY_PATH="$DIR/server/FileConverter/bin:${LD_LIBRARY_PATH:-}"
+
+            rm -f "$DIR"/sdkjs/*/sdk-all.cache
+            cd "$DIR/server/FileConverter/bin"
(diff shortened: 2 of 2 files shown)


chrip:
## Summary

The approach is sound and the mechanics are correct; the job wiring is not.

Sound approach: the no-ICU V8 that rejected the bundle in Euro-Office/sdkjs#80 only
exists inside the built image, so running it there is the one place the real
condition — "does this engine compile this bundle" — can actually be asserted. That
is strictly more than a source-tree check can do, and it catches the whole class
rather than the one known pattern.

Correct mechanics: the x2t invocation, the cache paths, the module list, the
container's user and the install paths all check out against `core` and the
Dockerfiles — I verified each one in the table below, including the
`LD_LIBRARY_PATH`/path question raised in the test plan.

The composite-action extraction is also well-timed: it was pulled out when a second
caller appeared, not speculatively on the first, and it is a strict no-op for `e2e`.

One blocking issue: the new job depends only on `build`, so on `main` and on tags it
resolves `:nightly` / `:latest` — tags that only the `manifest` job creates — and
therefore tests the *previous* run's image. That is the same "green check validated
the wrong artifact" failure that 373ee0c…d4f9b90 already fixed once for
`workflow_dispatch`; it survives in a second form on the default branch, which is
exactly where this gate matters most.

Note also that the job has not yet run: `build` is still in progress on this PR, so
the second test-plan checkbox is still unproven.

## Prior feedback

| From | Point | Status |
|------|-------|--------|
| @github-advanced-security (zizmor, inline `build.yml:538`) | code injection via template expansion — `${{ steps.pull.outputs.ref }}` interpolated into a `run:` script | ✓ addressed in 373ee0c (routed through `env: IMAGE_REF`); both zizmor checks are green at head |

No human reviews yet.

## Verification

| `x2t -create-js-cache` is a real flag and works with a single argument | `core/X2tConverter/src/main.cpp:152` — dispatched in the `argc == 2` path (`sArg1` is not `*.xml`, `sArg2` empty), calls `createJSCaches()` | ✓ |
| Caches land at `sdkjs/{word,cell,slide,visio}/sdk-all.cache` | `core/DesktopEditor/doctrenderer/editors.cpp:90,98,107,117` + `:135` appends `.cache` | ✓ |
| `<sdkjs>` config path resolves to `$DIR/sdkjs` | `build/configs/core/DoctRenderer.config` → `../../../sdkjs`, resolved from `NSFile::GetProcessDirectory()` (= `server/FileConverter/bin`) | ✓ |
| `/var/www/euro-office/documentserver/server/FileConverter/bin/x2t` still matches the image layout **(test-plan question)** | `server/Common/config/production-linux.json:74` names that exact path; `build/scripts/standalone/entrypoint.sh:7` sets `EO_ROOT` to the same | ✓ |
| `LD_LIBRARY_PATH` = `…/FileConverter/bin` is sufficient **(test-plan question)** | matches `document-server-package/common/documentserver/bin/documentserver-generate-allfonts.sh.m4:14`, the package's own convention for invoking these binaries | ✓ (see minor note below) |
| All four listed modules actually emit an `sdk-all.js` | `sdkjs/build/webpack.sdk.factory.mjs` builds word/cell/slide/visio, each emitting `sdk-all-min.js` + `sdk-all.js`; the VISIO branch's `Exists()` guard (`editors.cpp:112`) therefore passes | ✓ |
| `docker run --entrypoint bash` can `rm` and write into `sdkjs/` | `build/.docker/standalone.bake.Dockerfile` has no `USER` directive → runs as root; base is `ubuntu:24.04`, so `bash`/`stat` are present | ✓ |
| Exit code alone would catch Euro-Office/sdkjs#80 | No — `main.cpp:154-155` does `createJSCaches(); return 0;` unconditionally, so `rc` is 0 unless the process actually dies. The byte-size loop is the real assertion; keeping both is correct | ✓ (design is right) |
| `github.token` instead of `secrets.GITHUB_TOKEN` in the composite | Correct — the `secrets` context is not available inside composite actions; same value | ✓ |
| No stale `steps.image.outputs.ref` references left behind | Only `steps.pull.outputs.ref` remains (`build.yml:490,541`) | ✓ |
| "nothing in CI could have caught this" | Stale — `sdkjs/.github/workflows/check-build.yml:159` already runs `build/scripts/check-bundle-ascii.cjs` for exactly that failure mode | ❌ |
| `needs: build` is enough to have the right image | No — the bare `:nightly` / `:latest` tags are created by `manifest`, not `build` | ❌ |

## Issues & Suggestions

### 🔴 Blocking

- **The job can validate the previous run's image on `main` and on tags** —
  `.github/workflows/build.yml`, `sdkjs-v8-smoke` `needs:` / `if:`.
  `build` only pushes arch-suffixed tags (`nightly-${{ matrix.arch }}` /
  `latest-${{ matrix.arch }}`, meta step around `build.yml:70-77`). The bare
  `:nightly` and `:latest` refs that `pull-built-image` resolves to on non-PR events
  are created exclusively by the `manifest` job (`build.yml:395-440`, via
  `docker buildx imagetools create` from the `-amd64`/`-arm64` tags). Because
  `sdkjs-v8-smoke` needs only `build`, it starts concurrently with `manifest` and
  pulls whatever `:nightly` currently points at — the previous nightly — or races
  the manifest push nondeterministically. Result: on the default branch, the gate
  reports green about an artifact that is not the one just built. This is the same
  class of bug d4f9b90 fixed for `workflow_dispatch`, and `e2e` already guards
  against it — its condition requires `needs.manifest.result == 'success'` on the
  non-PR path (`build.yml:449-453`).
  Fix: `needs: [build, manifest]` and mirror `e2e`'s `if:` verbatim, including the
  `&& needs.manifest.result == 'success'` conjunct. Since the two conditions are now
  meant to be identical, they are also a good candidate for a shared job-level
  comment noting they must stay in sync.

### ⚠️ Major

- **A broken harness is reported as a broken bundle.** Dropping `set -e` was right
  for the diagnostic loop, but it means a wrong `$DIR`, a missing `x2t`, or an
  unresolvable shared library now takes the same path as a genuine V8 failure:
  `cd` fails, `./x2t` exits 127, and the annotation still reads
  `x2t -create-js-cache exited 127 — V8 likely aborted compiling sdk-all.js (see Euro-Office/sdkjs#80)`.
  Whoever hits that will go read the wrong issue. Add a precondition assert before
  the run, e.g.:
  ```bash
  [ -x "$DIR/server/FileConverter/bin/x2t" ] || {
    echo "::error::x2t not found at $DIR/server/FileConverter/bin — image layout changed, not a bundle problem"
    exit 2
  }
  ```
  and consider distinguishing `rc == 127` from other non-zero codes in the message.

- **The install path is brand-parametrised, not fixed.**
  `build/.docker/orchestrated.bake.Dockerfile:11-15,169` builds the tree under
  `/var/www/$COMPANY_NAME_LOW/documentserver` and exports `COMPANY_NAME_LOW` as an
  image `ENV`. Hardcoding `euro-office` is correct *today* only because the
  `nextcloud-office` brand is never pushed to `ghcr.io/euro-office` (meta step,
  `images_push=false`), so the pulled image is always the euro-office one. That is
  a non-obvious coupling to bake in silently. One line makes it self-correcting:
  ```bash
  DIR=/var/www/${COMPANY_NAME_LOW:-euro-office}/documentserver
  ```
  (with the guard, since `set -u` is on).

- **`runs-on: self-hosted` puts this on the same runner as `e2e`.** The workflow's
  own `concurrency` comment (`build.yml:42-46`) states that "the e2e job assumes a
  single runner". If the pool really is one runner, the description's "runs as its
  own job in parallel with e2e" does not hold — the two serialise, and the pipeline
  gets longer rather than the failure getting faster. This job needs only Docker and
  a GHCR pull of an amd64 image; `ubuntu-latest` would decouple it entirely and
  remove any contention with the Playwright run. How many self-hosted runners are
  registered?

### ℹ️ Minor

- **The stated justification is out of date.** Both the PR body and 003a7fc say
  `check-build.yml` "validates bundles with Node and headless Chromium only, so
  nothing in CI could have caught this". At the sdkjs pointer on this branch,
  `sdkjs/.github/workflows/check-build.yml:159` runs
  `node build/scripts/check-bundle-ascii.cjs` — `Assert built bundles are ASCII-only
  with no bare astral keys (#80)`. That does not make this job redundant: the ASCII
  check is a pattern-specific static proxy on the source tree, whereas this asserts
  the actual condition (does the real no-ICU V8 compile the bundle) on the actual
  shipped image, and generalises to the whole class. But the commit message should
  say *that* rather than claiming the gap is unguarded — it will read as wrong to
  anyone who greps sdkjs.

- **Bare `#80` in the PR description points at the wrong repository.** The "## Why"
  section opens with `#80: a Terser change emitted bare astral-plane object keys`.
  Posted on DocumentServer, `#80` autolinks to *this* repo's issue #80 — "Support
  Excel checkboxes" — not to Euro-Office/sdkjs#80. The summary bullet above it, the
  commit messages, the workflow comment and the `::error::` strings all use the
  fully-qualified `Euro-Office/sdkjs#80` and are fine; it is only that one reference.
  Worth fixing since the description is what a future reader lands on first.

- **Same interpolation pattern left in the composite's first step.** 373ee0c moved
  the pull step to `env:`, but `Determine image`
  (`.github/actions/pull-built-image/action.yml:30-38`) still expands
  `${{ inputs.registry }}`, `${{ inputs.cache-image-name }}` and
  `${{ github.event.pull_request.number }}` textually into `run:`. zizmor doesn't
  flag it (the PR number is an integer and the inputs are workflow-controlled), so
  this is consistency rather than a live vulnerability — but the commit claims the
  fix was applied to the action, and only half of it was.

- **License header:** none on the new `action.yml` — which is correct here. No file
  under `.github/` in this repo carries an SPDX header and there is no REUSE config,
  so adding one to a single new file would be the inconsistent choice. Flagging it
  explicitly so it isn't raised later as an oversight.

- `LD_LIBRARY_PATH` uses `…/FileConverter/bin`, matching
  `documentserver-generate-allfonts.sh.m4`. Note the supervisor units
  (`build/configs/standalone/supervisor/ds-converter.conf:5`) use
  `bin:lib`. If x2t ever fails to resolve a shared object here, appending
  `…/FileConverter/lib` is the fix — worth a comment so the next person doesn't
  have to rediscover it.

### 💡 Suggestions

- **Add `pdf` to the module loop — it's free coverage.**
  `CDoctrenderer::CreateCache` compiles WORD, SLIDE, CELL, VISIO *and* PDF
  (`core/DesktopEditor/doctrenderer/doctrenderer.cpp:1051-1055`), and PDF caches to
  `sdkjs/pdf/sdk-all.cache` (`editors.cpp:127`). The `rm -f "$DIR"/sdkjs/*/sdk-all.cache`
  glob already deletes it and x2t already regenerates it — it is simply not asserted.
  Adding it to `for module in word cell slide visio` also covers
  `pdf/src/engine/drawingfile_native.js` and `pdf/src/annotations/stamps.json`, which
  are appended only on the PDF path and are otherwise ungated.

- **`-eq 0` only catches "nothing written".** A truncated or partially written cache
  passes. A floor (`-lt 100000`, say) would catch that too, at the cost of one
  magic number.

- The checkout only needs `.github/actions/`; `sparse-checkout` would shave a bit
  off each run. Cosmetic.

## Code quality & conventions

- **Commits:** conventional and well-scoped — `refactor(ci):`, `ci:`, `fix(ci):` ×2.
  The refactor is correctly split from the feature, and both fix commits explain the
  *why* rather than the *what*. Good commit hygiene.
- **DCO:** `Signed-off-by:` present on all four commits; the DCO check is green.
- **AI disclosure:** `Assisted-by: ClaudeCode:claude-sonnet-5` on every commit and in
  the PR body. Compliant with AGENTS.md:102-104.
- **Scope:** focused. The composite-action extraction is justified by the second
  caller and is a strict no-op refactor for `e2e` (`github.token` ≡
  `secrets.GITHUB_TOKEN`; all references updated).
- **Tests:** n/a in the unit-test sense — the PR *is* a test. The relevant gap is
  that the job itself has not yet executed on this PR.
- **License headers:** consistent with the repo (see minor note above).

## Verdict

**Request changes** — the `needs`/`if` gap means this gate can go green against the
previous nightly on the default branch, which undermines the exact guarantee the PR
exists to provide; that one is a must-fix. The path/brand hardcoding and the
misleading abort diagnostic are cheap to fix while you're in there. Everything I
could check against `core`, the package Makefiles and the Dockerfiles holds up, and
the overall approach — asserting on the shipped image with the real no-ICU V8 rather
than on the source tree — is the right complement to the ASCII check already in
sdkjs. Happy to re-review once the job has actually run green here.

chrip (changes requested):
Following up on my review comment above (2026-08-28) — head is still d4f9b90a, so all
points there are unchanged. Re-posting as a formal review with two updates:

- ✓ `sdkjs-v8-smoke` has now run green (job 98859534243): word 5350008 / cell 6177376 /
  slide 4545144 / visio 4018792 bytes. That closes the second test-plan checkbox and
  confirms the image-layout and `LD_LIBRARY_PATH` assumptions at runtime.

- ❌ Withdrawing my `runs-on: self-hosted` concern — it was wrong. This job and `e2e`
  ran on two distinct ephemeral runners (`garm5-24-eo-bk3dlpvmrspq` /
  `garm5-24-eo-tef3csdt8gkt`), both starting 13:14:35Z, genuinely in parallel. Leave
  it as `self-hosted`.

Still blocking: `sdkjs-v8-smoke` needs only `build` (build.yml:512-513), so on `main`
and tags it resolves `:nightly` / `:latest` — tags only `manifest` creates — and can
test the previous run's image. The green run here doesn't cover it: `manifest` is
skipped on `pull_request`, so this run used `:pr-350` and never took that path. Fix is
`needs: [build, manifest]` plus e2e's `if:` verbatim, including
`&& needs.manifest.result == 'success'`.

The two ⚠️ hardening points (precondition assert on `$DIR`/`x2t` before the run, and
`${COMPANY_NAME_LOW:-euro-office}` instead of the hardcoded path) and the minor/💡
items are cheap to do while you're in there.

PR #351, opened 2026-08-29
Fix candidate list formatting in auto-ping workflow

renatoyamane:
Updated usernames list


DmySyz:
Hi, done in a different PR and removed an additional outdated triager. Thank you for bringing this up!

Issue #44, opened 2026-04-01
How will you solve the OnlyOffice license violation?


Simple-Minute:
Hello,
how will Euro-Office solve the violation of OnlyOffice license?
https://www.onlyoffice.com/blog/2026/03/onlyoffice-flags-license-violations-in-euro-office-project-by-nextcloud-and-ionos

I could find similar case being tried in US, Neo4j v. PureThink.
Its a case where Neo4j used "AGPL + additional restrictions" license and PureThink tried to redistribute the product with the restrictions removed.
There court decided you cant just remove "additional restrictions" from the license.

> That's the court saying the AGPL can be modified with the addition of the Commons Clause license and that modification cannot be undone, despite what the AGPL says.
https://www.theregister.com/2025/02/27/adverse_appeals_court_ruling_could/

> The actually interesting part of this whole kerfuffle is that the trial court decided PureThink could not rip Commons Clause out of the LICENSE file

https://writing.kemitchell.com/2022/03/17/OSI-Neo4j-PureThink

While this case in in US and not EU, its the only similar decision I could find.
So better to think about it.

tiran133:
> Because there is a claim and I (to my best knowledge) don't think there is an official statement from the project owners.

Exactly my point. Why is there no official statement?

mnaljm:
> *Phase 2: Your Defense (The "Sega" Strategy)*

Did you post a chat from your conversation with your AI about a fictious court case about this?

allenwrench218:
its not fictitious, ive done my homework
https://en.wikipedia.org/wiki/Sega_v._Accolade

ozantokman:
Why in the codebase still there is " onlyoffice " folder. Isn't that strange ?

allenwrench218:
> Why in the codebase still there is " onlyoffice " folder. Isn't that strange ?

same reason libreoffice's binary is called soffice, for compatibility. Still this license is a nonissue, what they say and what is codified in the license means either way eurooffice will win, because if they move the logo to the credits and strict verbatim interpretation the court will strike that clause and if theres a conflict theres a get out of jail free near the end of section 7.

chriscroome:
Adding a link to the FSF article on this issue:

https://www.fsf.org/blogs/licensing/agpl-is-not-a-tool-for-taking-freedom-away

[...]

PR #363, opened 2026-09-10
fix(entrypoint): make a Redis-backed editorData selectable, and stop fabricating a sentinel

moodyjmz:
Companion to `Euro-Office/server#44`. Neither is much use without the other: that PR adds a Redis-backed `editorData` store, this one makes it possible to configure and select on an orchestrated image. Draft until #44 is ready.

Four commits, each independently reviewable.

## 1. Stop listing the Redis server as its own sentinel

When `REDIS_SENTINEL_NODES` is empty the `else` branch emitted a `sentinels` array containing the plain Redis server itself, and `iooptions.sentinels` was written unconditionally. A populated `sentinels` array is exactly what selects sentinel mode in ioredis, so any consumer spreading `iooptions` into a client entered sentinel mode on a deployment with no sentinel, and never connected.

An enterprise deployment hit this in production: the client sat in a permanent retry loop at ~80 sockets per pod with `All sentinels are unreachable ... ERR unknown command 'sentinel'`, and because that consumer's locks fail closed, **every save was refused while the deployment reported healthy** — 0 of 461 saves acknowledged, nothing in the log.

The `"sentinels":` prefix moves into the variable, because blanking a value behind a hardcoded key leaves invalid JSON. A deployment that genuinely configures sentinels is unaffected.

## 2. State the topology as `services.CoAuthoring.redis.mode`

Derived from variables the deployment already sets — cluster when `REDIS_CLUSTER_NODES` is present, sentinel when `REDIS_SENTINEL_NODES` is, standalone otherwise — with `REDIS_MODE` overriding. A consumer previously had to infer topology from the *shape* of the options, and that inference is what commit 1 is made of.

## 3. Select the backend with `EDITOR_DATA_STORAGE`

The entrypoint reads twelve `REDIS_*` variables but exposed no way to actually select a Redis-backed `editorData` store. Since it exports `NODE_CONFIG`, which outranks every config file, `local.json` cannot do it either — the only route was a `--NODE_CONFIG` command-line argument, itself needing a bare `--` because `getopts` runs over what follows.

`EDITOR_DATA_STORAGE` and `EDITOR_STAT_STORAGE` are emitted only when set, so a deployment that says nothing keeps whatever the package shipped. The default stays memory deliberately: a Redis-backed store whose locks fail closed would refuse every save wherever Redis is absent, behind a healthy healthcheck.

## 4. Validate what reaches `require()` and the config

These three values are interpolated into `NODE_CONFIG` unescaped, like every other value here, and the two storage names then reach `require('./' + value)` in docservice. A mode outside the known set, or a storage name that is not a plain module basename, is now rejected before interpolation.

Not a new trust boundary — anyone who can set container environment can already supply a full `NODE_CONFIG` — but cheap to constrain, and the error names the variable, which connection-refused later does not.

## Verification

`bash -n` clean. Run inside the real image, under the image's own bash, sourcing the script's own variable logic: valid `NODE_CONFIG` emitted across standalone, genuine-sentinel, cluster and explicit-`REDIS_MODE`, and across all four combinations of the two storage variables — absent entirely when neither is set. A bad `REDIS_MODE` and a path-injection attempt in `EDITOR_DATA_STORAGE` are both rejected with a message naming the variable.

## Risk

Commits 2–4 are inert on their own: nothing in the current tree reads `redis.mode`, and the storage keys only change behaviour for a deployment that sets them. Commit 1 changes emitted config for every non-sentinel deployment, but nothing in the tree reads `iooptions` today, so no current deployment changes behaviour.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/build/scripts/orchestrated/docker-entrypoint.sh
+++ b/build/scripts/orchestrated/docker-entrypoint.sh
@@ -62,6 +62,18 @@ else
   REDIS_SENTINEL=''
 fi
 
+# Only name a user when one was configured. ioredis sends the two-argument
+# AUTH whenever a username is present - a password is not required for that -
+# and two-argument AUTH is Redis 6.0 and later. Emitting a default username
+# unconditionally therefore made 6.0 the floor for every deployment; on Redis
+# 5 with a password set it fails, warns, and leaves the client
+# unauthenticated for every subsequent command.
+if [[ -n "$REDIS_SERVER_USER" ]]; then
+  REDIS_USERNAME_JSON='"username": "'${REDIS_SERVER_USER}'",'
+else
+  REDIS_USERNAME_JSON=''
+fi
+
 if [[ -n "$REDIS_CLUSTER_NODES" ]]; then
   declare -a REDIS_CLUSTER_NODES_ALL=($REDIS_CLUSTER_NODES)
   REDIS_CLUSTER_NODES_ARRAY=()
@@ -72,7 +84,7 @@ if [[ -n "$REDIS_CLUSTER_NODES" ]]; then
   IFS=","
   NODES=$(echo "${REDIS_CLUSTER_NODES_ARRAY[*]}")
   IFS="$OLD_IFS"
-  REDIS_CLUSTER='"rootNodes": [ '$NODES' ], "defaults": { "username": "'${REDIS_SERVER_USER:-default}'", "password": "'$REDIS_SERVER_PWD'" }'
+  REDIS_CLUSTER='"rootNodes": [ '$NODES' ], "defaults": { '${REDIS_USERNAME_JSON}'"password": "'$REDIS_SERVER_PWD'" }'
 else
   REDIS_CLUSTER=''
 fi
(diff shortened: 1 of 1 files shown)


moodyjmz:
## The two red `nextcloud-office` builds are pre-existing, not from this PR

`build (amd64|arm64, nextcloud-office, …)` fail here, while both `euro-office` builds pass on the same commit.

```
ERROR: resolve : lstat ../../../build: no such file or directory
##[error]buildx bake failed with: ERROR: resolve : lstat ../../../build: no such file or directory
```

It fails in `Build all targets (packages → docker → develop)`, while `buildx bake` is resolving a context — before anything in this PR is executed.

### Why it cannot be this PR

- This PR changes one file: `build/scripts/orchestrated/docker-entrypoint.sh`. Both brandings ship **that same script**, so a fault in it would fail both, not one.
- The failure is a bake context resolution, which happens before any entrypoint runs.
- The identical pattern — both `nextcloud-office` jobs red, both `euro-office` jobs green — appears on `chore/nightly-submodule-bump` (run 34466459886), which touches no entrypoint at all.

### What it looks like

The only difference between the two matrix entries is the branding overlay:

| | `euro-office` | `nextcloud-office` |
|---|---|---|
| `extra_bake_files` | `""` | `./brands/nextcloud-office-brand/brand-server.hcl` |
| `extra_targets` | `""` | `brand-icons,` |

`brand-server.hcl` comes from `Euro-Office/nextcloud-office-brand`, checked out into `build/brands/` in CI only, so it is not in a normal working tree. A relative path in that overlay is not resolving from wherever bake anchors it — `../../../build` lands outside the workspace. The unbranded build passes because it layers no second bake file and so has no such path to resolve.

I have not opened the overlay to find the offending line — it is a second unrelated pre-existing break and I did not want to chase it inside this PR.

@rikled — `git blame` puts the branding matrix and `extra_bake_files` on you (d7eab9ba), so you are likely the fastest person to confirm whether that relative context is meant to be anchored differently. No urgency from our side; flagging it because it is red on every branch that runs this workflow, not just this one.

Happy to raise it as its own issue with this diagnosis if that is more useful than a comment here.

